PR #9, opened 2015-06-09
Remove unused AES primitives functions and exports

tekul:
The code for initializing different AEAD modes is now encapsulated in
the BlockCipher type and the individual mode encryption and decryption
functions have been replaced by generalized versions, so are no longer
used.

I think we talked briefly about this. You might have some other motive for keeping the code, in which case, feel free to ignore this.


vincenthz:
great, thanks !

PR #20, opened 2015-06-29
Use non-blocking IO with /dev/random.

rblaze:
I'm adding Network.TLS as HTTPS provider to happstack and run into the problem on Linux VM in Azure. First request is served instantly, but every next one is blocked for up to 20 seconds. Quick scan with strace showed that process is waiting for at least one byte to be available from the /dev/random.
This is quick hack to skip /dev/random if it is not ready. I also wonder if it is expected for entropy pool to be refreshed on every handshake, it would hurt the performance.


vincenthz:
Looks like a mix up, it was supposed to bail if there was no data available in random, thanks for the patch definitely a good thing ! I think on Linux `random` should be skipped altogether, but on some other unix platform it might be different (hence why it's there)

PR #41, opened 2015-10-27
Fix typo: Kekkak -> Keccak

xtendo-org:
First of all, thank you Vincent! I have benefited multiple times from your effort. Your contribution to Haskell's cryptography and security code base is incredible.

Hopefully I'm not much mistaken and this pull request is useful. The SHA-3 winning algorithm's name is Keccak (pronounced "catch-ack"), not Kekkak.
- http://keccak.noekeon.org/
- http://www.nist.gov/itl/csd/sha-100212.cfm
- https://en.wikipedia.org/wiki/SHA-3

Obviously, this is a breaking change to the API.


vincenthz:
oh. good catch, this is useful.

Not sure this is useful, as not that many people should be using the `Keccak` types compared to `SHA3`, but maybe it make sense to provide the same `Kekkak` types that use the `Keccak` implementation.

vincenthz:
merging without any compatibility layer, probably good enough like this. Thanks !

Issue #44, opened 2015-11-10
Can't implement hash functions


centromere:
I'd like to use the HMAC functionality of cryptonite with an unsupported hash function (BLAKE2b/s), but I am unable to do so because the HashAlgorithm class is not exported.

vincenthz:
I'ld like to think it's by design, so that someone would contribute the BLAKE2s/b implementation :)

There should be almost everything needed to generate the haskell bits, (in the `gen` directory), so it's "just" missing the C implementation in cbits/

centromere:
The entries in Gen.hs look like this:

```
, GenHashModule "SHA512"    "sha512.h"    "sha512"    256 64 128 []
```

BLAKE2 has a variable output length. What's the best way to proceed?

vincenthz:
I haven't look at the specific, but I would assume just like `SHA3` or `Skein` that have variable output length, the best is to specialize it to whichever length are commonly used.

centromere:
This is no longer relevant to me.

PR #45, opened 2015-11-12
Added BLAKE2 support

centromere:
#44


vincenthz:
Looking good ! 2 things before merging this:
- add an item in the CHANGELOG
- make sure there's no C functions exported without the cryptonite_ prefix, that include all the blake2\* symbols, to prevent collision with other packages/libraries/etc

centromere:
Done.

vincenthz:
excelllent ! Thanks

PR #46, opened 2015-11-15
Fix endianness of incrementNonce function for ChaChaPoly1305

centromere:
This also fixes an issue where 12-byte nonces are created in the nonce8 function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Add missing documentation for various modules
 * Add a way to create Poly1305 Auth tag.
 * Added support for the BLAKE2 family of hash algorithms
+* Fix endianness of incrementNonce function for ChaChaPoly1305
 
 ## 0.8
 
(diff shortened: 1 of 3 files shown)


vincenthz:
Why did you remove the constant when creating the nonce8 ? this is needed for 64 bits nonce generator like in TLS and backed by rfc7539:

```
 ChaCha20 as specified here requires a 96-bit
 nonce.  So if the provided nonce is only 64-bit, then the first 32
 bits of the nonce will be set to a constant number.  This will
 usually be zero, but for protocols with multiple senders it may be
 different for each sender, but should be the same for all
 invocations of the function with the same key by a particular
 sender.
```

centromere:
You're absolutely right, I screwed up. My intention was to have `incrementNonce` handle both 12-byte and 8-byte nonces properly. I believe I've corrected this error in the commit above. What do you think?

vincenthz:
Looking much better ! I think it's probably "more correct" to separate the cases, and have the increment work over the 8 bytes "domain" in the case of the nonce8 case.

PR #51, opened 2015-11-30
Add support_blake2 flag.

mpietrzak:
Blake2 does not compile on arm.

Adding this flag makes it possible to compile this library on arm arch and by extension makes it possible to compile and run Yesod apps on arm systems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10
+
+* Add support_blake2 flag
+
 ## 0.9
 
 * Quiet down unused module imports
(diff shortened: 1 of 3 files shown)


centromere:
Can you add a line to the change log?

mpietrzak:
@centromere added

centromere:
@vincenthz Do you approve?

vincenthz:
Yes that's probably fine for now. I'll try to fix it to not require SSE2 and fail gracefully, in the future; it seems possible.

mpietrzak:
Thanks!
Note that the flag is by default true and will not turn itself off automatically.
So by default compilation will still fail on e.g. arm systems.
I didn't find a way for flag default value to depend on arch, putting default value in `if` didn't work for me.
Maybe we should also turn Blake2 off on arm automatically (for now, while it still requires SSE)? That is replace `if flag(support_blake2)` with `if flag(support_blake2) && arch(arm)`?
Anyway, the flag is meant as a small improvement - certainly not complete fix, but something that at least makes compilation (of the rest of package) possible on arm.

PR #53, opened 2015-12-03
[ChaChaPoly1305] fix type error of example code

xtendo-org:
The example code had a type mismatch.

```
Couldn't match expected type ‘State’
            with actual type ‘CryptoFailable State’
In the second argument of ‘appendAAD’, namely ‘st1’
In the second argument of ‘($)’, namely ‘appendAAD hdr st1’
```

This is due to the following part:

```
let st1 = ChaChaPoly1305.initialize key nonce
    st2 = ChaChaPoly1305.finalizeAAD $ ChaChaPoly1305.appendAAD hdr st1
```

`initialize` returns `CryptoFailable State`, not `State`.

This commit fixes the type mismatch, changes the return type of the example function to `CryptoFailable ByteString`, and makes the code to be immediately copy-and-paste-able.

By the way, will it be a good idea to put the corresponding `decrypt` function? (having the type of `CryptoFailable ByteString` with a correct auth tag checking.)

### DIFF
--- a/Crypto/Cipher/ChaChaPoly1305.hs
+++ b/Crypto/Cipher/ChaChaPoly1305.hs
@@ -5,7 +5,8 @@
 -- Stability   : stable
 -- Portability : good
 --
--- A simple AEAD scheme using ChaCha20 and Poly1305. See RFC7539.
+-- A simple AEAD scheme using ChaCha20 and Poly1305. See
+-- <https://tools.ietf.org/html/rfc7539 RFC 7539>.
 --
 -- The State is not modified in place, so each function changing the State,
 -- returns a new State.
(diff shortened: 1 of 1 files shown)


xtendo-org:
Another (maybe related, maybe orthogonal) question:
- [RFC 7539](https://tools.ietf.org/html/rfc7539)
- [Draft for ChaCha20-Poly1305 in TLS](https://tools.ietf.org/html/draft-ietf-tls-chacha20-poly1305-02)

Both of them refers to the scheme with the name "ChaCha20-Poly1305." The current module name is `ChaChaPoly1305`.

So, "20" seems to be missing. Is this intended?

vincenthz:
Thanks.

For the missing 20, it's a simple omission but unless there's some conflicts in what the name means, I suspect it's a bit too costly to fix now; However, if someone want to fix this, one could create a duplicated ChaCha20Poly1305 module and add a deprecation to the current one in several months or so.

Issue #55, opened 2015-12-15
build failed on ppc/ppc64/ppc64le/aarch64


mimi1vx:
112s] [99 of 99] Compiling Crypto.Cipher.AES ( Crypto/Cipher/AES.hs, dist/build/Crypto/Cipher/AES.p_o )
[  139s] 
[  139s] In file included from cbits/blake2/blake2s.c:21:0: 
[  139s] 
[  139s] cbits/blake2/blake2-config.h:68:2:
[  139s]      error: #error "This code requires at least SSE2."
[  139s]      #error "This code requires at least SSE2."
[  139s]       ^
[  139s] 
[  139s] cbits/blake2/blake2s.c:24:23:
[  139s]      fatal error: emmintrin.h: No such file or directory
[  139s] compilation terminated.

centromere:
@mimi1vx Does the `blake2-fix` branch build for you?

mimi1vx:
@centromere with blake2-fix works without any observable problems .. see https://build.opensuse.org/package/show/devel:languages:haskell:lts/ghc-cryptonite

centromere:
@katsuyas @mpietrzak Does the `blake2-fix` branch work in your environment? No flags should need to be specified.

mimi1vx:
@centromere for openSUSE packaging must be disabled AESNI and sse-4.1 +rdrnad :(

mpietrzak:
It does compile, I'm not sure yet if it works.
Is there an easy way to check if it works? I'm compiling tests now, but it takes a long time.

btw I'm using https://www.scaleway.com/

centromere:
@mimi1vx Ah, my statement was unclear. I meant to say, `support_blake2_sse` should be disabled by default, so you shouldn't need to specify it one way or the other.

bgamari:
I am also seeing this on ARM. The `blake2-fix` branch appears to build without any trouble.

centromere:
@vincenthz For your consideration. ^

wamserma:
Same as @bgamari - ARM does not have SSE2. cryptonite happens to be in the dependency tree of hakyll, so having this build break is very annoying. Having the fix on hackage so i can just cabal update would be very appreciated as a Christmas present. :)

centromere:
@wamserma According to others, the above fix seems to be working. I can merge it to master, but only @vincenthz can push to hackage.

wamserma:
[...]

Issue #54, opened 2015-12-15
cryptonite-0.9 fails to build on i586


mimi1vx:
from build log:

  149s] [99 of 99] Compiling Crypto.Cipher.AES ( Crypto/Cipher/AES.hs, dist/build/Crypto/Cipher/AES.p_o )
[  185s] 
[  185s] In file included from cbits/blake2/blake2s.c:21:0: 
[  185s] 
[  185s] cbits/blake2/blake2-config.h:68:2:
[  185s]      error: #error "This code requires at least SSE2."
[  185s]      #error "This code requires at least SSE2."
[  185s]       ^
[  185s] 
[  185s] In file included from cbits/blake2/blake2s.c:24:0: 
[  185s] 
[  185s] /usr/lib/gcc/i586-suse-linux/4.8/include/emmintrin.h:31:3:
[  185s]      error: #error "SSE2 instruction set not enabled"
[  185s]      # error "SSE2 instruction set not enabled"
[  185s]        ^
[  185s] cbits/blake2/blake2s.c: In function ‘blake2s_compress’:
[  185s] 
[  185s] cbits/blake2/blake2s.c:242:3:
[  185s]      error: unknown type name ‘__m128i’
[  185s]        __m128i row1, row2, row3, row4;
[  185s]        ^
[  185s] 
[  185s] cbits/blake2/blake2s.c:243:3:
[  185s]      error: unknown type name ‘__m128i’
[  185s]        __m128i buf1, buf2, buf3, buf4;
[  185s]        ^
[  185s] 
[  185s] cbits/blake2/blake2s.c:250:3:
[  185s]      error: unknown type name ‘__m128i’
[  185s]        __m128i ff0, ff1;
[  185s]        ^
[  185s] 
[  185s] cbits/blake2/blake2s.c:278:3:
[  185s]      warning: implicit declaration of function ‘_mm_loadu_si128’ [-Wimplicit-function-declaration]
[  185s]        row1 = ff0 = LOADU( &amp;S-&gt;h[0] );
[  185s]        ^
[  185s] 
[  185s] cbits/blake2/blake2s.c:278:3:  error: unknown type name ‘__m128i’
[  185s] 
[  185s] cbits/blake2/blake2s.c:279:3:
[  185s]      error: unknown type name ‘__m128i’
[  185s]        row2 = ff1 = LOADU( &amp;S-&gt;h[4] );
[  185s]        ^
[  185s] 
[...]

PR #56, opened 2015-12-21
[blake2] Added reference implementation

centromere:
This commit allows the user to select either the portable reference
implementation or the optimized (SSE) implementation.

See #54, #55.


vincenthz:
awesome work. merry Yule.

PR #63, opened 2016-01-11
[For Review] HOTP and TOTP implementation

tekul:
This is the current implementation of the HOTP and TOTP specs I've been working on.

Any comments welcome, particularly on the API, use of config parameters etc.

### DIFF
--- a/Crypto/OTP.hs
+++ b/Crypto/OTP.hs
@@ -11,10 +11,27 @@
 --
 -- Probably the best-known use of TOTP is in Google's 2-factor authentication.
 --
+-- The TOTP API doesn't depend on any particular time package, so the user needs to supply
+-- the current @OTPTime@ value, based on the system time. For example, using the @hourglass@
+-- package, you could create a @getOTPTime@ function:
+--
+-- >>> import Time.System
+-- >>> import Time.Types
+-- >>>
+-- >>> let getOTPTime = timeCurrent >>= \(Elapsed t) -> return (fromIntegral t :: OTPTime)
+--
+-- Or if you prefer, the @time@ package could be used:
+--
+-- >>> import Data.Time.Clock.POSIX
+-- >>>
+-- >>> let getOTPTime = getPOSIXTime >>= \t -> return (floor t :: OTPTime)
+--
 
 module Crypto.OTP
-    ( hotp
+    ( OTP
     , OTPDigits (..)
+    , OTPTime
+    , hotp
     , resynchronize
     , totp
     , totpVerify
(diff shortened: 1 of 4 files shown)


Crypto/OTP.hs
```diff
@@ -0,0 +1,168 @@
...
+    -> Bool
+totpVerify (TP h t0 x d skew) k now otp = otp `elem` map (hotp h d k) (range window [])
+  where
+    t = timeToCounter now t0 x
+    window = fromIntegral (fromEnum skew)
+    range 0 acc = t : acc
+    range n acc = range (n-1) ((t-n) : (t+n) : acc)
+
+timeToCounter :: POSIXTime -> Word64 -> Word16 -> Word64
+timeToCounter now t0 x = floor ((now - fromIntegral t0) / fromIntegral x)
```
vincenthz: POSIXTime is not a good representation. it has a 10^-12 representation that is useless here and I'm not a fan to depends on `time`, I think it would make more sense to just:

```
type OTPTime = Word64
or
type SecondSinceUnix = Word64
```

tekul: It's true the extra resolution is no use and it's a pain to add a dependency just for this type, but presumably the end user will only be able to call `getPOSIXTime` or `fmap utcTimeToPOSIXSeconds getCurrentTime` to get the current time and would end up using this type in any case (they'd then have to convert). I don't know of any other functions which do the same thing. I guess we could add a function of type `IO OTPTime`  implemented in a similar way to these, but without needing the `time` dependency.

Crypto/OTP.hs
```diff
@@ -0,0 +1,168 @@
...
+defaultTOTPParams :: TOTPParams SHA1
+defaultTOTPParams = TP SHA1 0 30 OTP6 TwoSteps
+
+-- | Create a TOTP configuration with customized parameters.
+mkTOTPParams :: (HashAlgorithm hash)
+    => hash
+    -> Word64
+    -- ^ The T0 parameter in seconds. This is the Unix time from which to start
+    -- counting steps (default 0). Must be before the current time.
+    -> Word16
```
vincenthz, line 138: Int here maybe ? This is probably better as a more neutral Int than Word16, specially since it doesn't cover the whole Word16 range.

tekul, line 138: Not sure what you mean - Word16 has a much smaller range than Int. If I used Int I'd also have to check for and reject negative values, so it seemed best to indicate in the type that it has to be positive.

Crypto/OTP.hs
vincenthz: They could (I would :-)) use `timeCurrent` from `hourglass` as well. I'm not sure it's a good idea to have a IO callback, as this would make the function impure, and un-replayable (if you want to test known values or passed/future values).

tekul: Hmm, they could, yes. I forgot about that library :-). I meant just adding a function like `timeCurrent :: IO OTPTime` within the module itself, rather than modifying the function types as already written.

vincenthz:
It's probably be a good idea to move forward with this; I think my only requirements is dropping `time` before merging. I'm fine with a callback to get the time in the necessary shape to prevent a hard dependency on time.

tekul:
My idea was just to re-implement a time function internally which a user could call for convenience, but that would also require a dependency on Win32 to support windows, which would also be a pain. I guess I'll just stick with a Word64 type in the API and let the user create the values as they wish.

vincenthz:
I think the Word64 is the simplest and best option. We can also have a comment above the function on how to get the required value with `time` ... and/or `hourglass` :-)

centromere:
@vincenthz Now that `time` has been dropped, can we merge this pull request?

tekul:
I think we were going to do it at Zurihac. Maybe we meant Zurihac 2017.

vincenthz:
yay ~7 months before zurihac 2017, what a speedy reviewer (!)

Looking great, well documented and nicely typed, sorry about the delay, absolutely no reason to hold it longer

PR #68, opened 2016-02-17
Use a build hook to define the build configuration

vincenthz:
I still absolutely hate the approach as this cause all sort of cabal bugs to come out, but I think it's becoming necessary considering all the build problem with the broad range of system supported.

@centromere @tekul any opinions ?

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -1,2 +1,95 @@
+-- CHANGELOG since copy:
+-- * refactor to allow more check than AESNI support
+--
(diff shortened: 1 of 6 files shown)


vincenthz:
maybe @tommd is interested in this too.

tekul:
I don't know enough about cabal to have an opinion. Hopefully I can keep it that way :).

vincenthz:
As feared, the tests were a train wreck .. so I think we can leave this for future considerations if need arise again, but for now carry on with the simple setup

Issue #74, opened 2016-04-01
Adding CMAC support


khibino:
I want to use CMAC (Cipher based Message Authenticaton Code) with cryptonite.
I implemented a naive implementation and test vectors.

May I create a pull request for this implementation?

https://github.com/khibino/cryptonite/commit/ca0c3830ebad0eb61bb07344516a8e7fd272be03
https://github.com/khibino/cryptonite/commit/b704f2c02ad4d66395ec24b5eedc2d32bc83c6fd

>    CMAC
>      KAT - NIST test vectors
>        aes128
>          1:                 OK
>          2:                 OK
>          3:                 OK
>          4:                 OK
>          5:                 OK
>          6:                 OK
>          7:                 OK
>        aes192
>          1:                 OK
>          2:                 OK
>          3:                 OK
>          4:                 OK
>          5:                 OK
>          6:                 OK
>          7:                 OK
>        aes256
>          1:                 OK
>          2:                 OK
>          3:                 OK
>          4:                 OK
>          5:                 OK
>          6:                 OK
>          7:                 OK
>        Three Key TDEA
>          1:                 OK
>          2:                 OK
>          3:                 OK
>          4:                 OK
>          5:                 OK
>          6:                 OK (0.01s)
>          7:                 OK (0.01s)
>        Two Key TDEA
>          1:                 OK
>          2:                 OK
>          3:                 OK
>          4:                 OK
>          5:                 OK
>          6:                 OK (0.01s)
>          7:                 OK (0.01s)

vincenthz:
yes, please.

as a pre-review, it's looking good. the only thing that strikes me is that the MAC generated should be a type not just a ByteArray constraint. Maybe something like:

```
newtype CMAC = CMAC Bytes
```

This way you can add a time equality constant function on the `Eq` instance, just like for HMAC:

```
instance Eq (HMAC a) where
    (HMAC b1) == (HMAC b2) = B.constEq b1 b2
```

khibino:
I added CMAC newtype with cipher phantom type, and created its pull request.
https://github.com/haskell-crypto/cryptonite/pull/75

PR #75, opened 2016-04-06
Adding CMAC support

khibino:
Pre-reviewed in https://github.com/haskell-crypto/cryptonite/issues/74

### DIFF
--- a/Crypto/MAC/CMAC.hs
+++ b/Crypto/MAC/CMAC.hs
@@ -1,0 +1,117 @@
+-- |
+-- Module      : Crypto.MAC.CMAC
+-- License     : BSD-style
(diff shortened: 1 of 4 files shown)


vincenthz:
Thanks, very nice

vincenthz:
@khibino Could you check if you're happy with a73c1b91718f3b2bd4fd59022d992da7ff621b8e ? I removed the accessor and the direct access to CMAC in favor of the ByteArrayAccess type class

khibino:
> @khibino Could you check if you're happy with a73c1b9 ? I removed the accessor and the direct access to CMAC in favor of the ByteArrayAccess type class

No, problem. Thanks for merging.

PR #77, opened 2016-04-12
Adding miyaguchi-preneel hash construction

khibino:
Hi.

I want miyaguchi-preneel hash construction
(known as one-way compression function
 https://en.wikipedia.org/wiki/One-way_compression_function#Miyaguchi.E2.80.93Preneel)
available with cryptonite block-cipher.

I implemented a naive implementation and test vectors.

### DIFF
--- a/Crypto/ConstructHash/MiyaguchiPreneel.hs
+++ b/Crypto/ConstructHash/MiyaguchiPreneel.hs
@@ -10,7 +10,7 @@
 --
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 module Crypto.ConstructHash.MiyaguchiPreneel
-       ( mp
+       ( mp, mp'
        , MiyaguchiPreneel(..)
        , cipherInit'
        ) where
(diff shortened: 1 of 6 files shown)


Crypto/ConstructHash/MiyaguchiPreneel.hs
```diff
@@ -0,0 +1,76 @@
...
+    bsz = blockSize ( g B.empty {- dummy to get block size -} )
+    chunks msg
+      | B.null tl  =  [hd :: Bytes]
+      | otherwise  =   hd : chunks tl
+      where
+        (hd, tl) = B.splitAt bsz msg
+
+-- | Simple key build function, which may raise size error.
+cipherInit' :: (ByteArray ba, Cipher k) => ba -> k
+cipherInit' = either (error . show) id . eitherCryptoError . cipherInit
```
vincenthz: `either error ... . eitherCryptoError` is the same as `throwCryptoError`.

Also `cipherInit'` is not actually dependent of MiyaguchiPreneel construction, so I'ld rather just delete this function from this module.

Crypto/ConstructHash/MiyaguchiPreneel.hs
```diff
@@ -0,0 +1,76 @@
...
+     -> ba
+     -> ba
+step g iv msg =
+    ecbEncrypt k pmsg `bxor` iv `bxor` pmsg
+  where
+    k = g iv
+    pmsg = pad0 k msg
+
+pad0 :: (ByteArray ba, BlockCipher k) => k -> ba -> ba
+pad0 k s = s `B.append` B.replicate (blockSize k - B.length s) 0
```
vincenthz: Probably a good idea to move this in the Padding module

tests/KAT_MiyaguchiPreneel.hs
```diff
@@ -0,0 +1,53 @@
...
+import           Data.Char (digitToInt)
+import qualified Data.ByteString as BS
+import qualified Data.ByteArray as B
+
+
+runMP128 :: ByteString -> ByteString
+runMP128 s = B.convert (mp s :: MiyaguchiPreneel AES128)
+
+hxs :: String -> ByteString
+hxs = BS.pack . rec' where
```
vincenthz: there's a Data.ByteArray.Encoding module that gives you Base16 decoding/encoding for free.

Crypto/ConstructHash/MiyaguchiPreneel.hs
```diff
@@ -0,0 +1,76 @@
...
+
+newtype MiyaguchiPreneel a = MP Bytes
+    deriving (ByteArrayAccess)
+
+instance Eq (MiyaguchiPreneel a) where
+    MP b1 == MP b2  =  B.constEq b1 b2
+
+
+-- | Compute Miyaguchi-Preneel one way compress using the supplied block cipher.
+mp' :: (ByteArrayAccess bin, BlockCipher cipher)
```
vincenthz: I don't think `mp'` is a good name for this, specially as this is the exported function. What about something more self-documenting like `miyaguchiPreneel`, or provided you advise to use the module qualified something like `compute` (my preferred solution)

vincenthz:
Generally looks good; I think a new hierarchy is a good idea, and this looks almost ready. Could you fixed the little niggles I noted in the inlined comments ? thanks

vincenthz:
Thanks. looking good now

PR #78, opened 2016-04-13
more instances for (Digest a)

paolino:
implementing Merkle root I needed 
`hash (x <> y)`

### DIFF
--- a/cryptonite.cabal
+++ b/cryptonite.cabal
@@ -192,7 +192,7 @@ Library
                    , bytestring
                    , memory >= 0.8
                    , ghc-prim
-  ghc-options:       -Wall -fwarn-tabs -optc-O3 -fno-warn-unused-imports
+  ghc-options:       -Wall -fwarn-tabs -optc-O3 -fno-warn-unused-imports -fobject-code
   default-language:  Haskell2010
   cc-options:        -std=gnu99
   if flag(old_toolchain_inliner)
--- a/Crypto/Hash.hs
+++ b/Crypto/Hash.hs
@@ -104,8 +104,4 @@ hashInitWith _ = hashInit
 hashWith :: (ByteArrayAccess ba, HashAlgorithm alg) => alg -> ba -> Digest alg
 hashWith _ = hash
 
-instance HashAlgorithm a => Monoid (Digest a) where
-    mempty = hashFinalize hashInit
-    mappend x y = hashFinalize $ foldl' hashUpdate hashInit [x,y]
-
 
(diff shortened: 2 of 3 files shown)


paolino:
Read instance for `Digest a` in Base16 as Show instance.
read check both encoding and length.
I had to move `digestFromByteString` to Hash.Types, but it's still exported from Hash

vincenthz:
@paolino: this can't be right; this allow creation of, `Digest a` which could be of any size, which break the assumption that it represent a digest for the algorithm `a`

Why not simply use `hashFinalize $ foldl' hashUpdate hashInit [x,y]` ?

paolino:
You are right. If you agree that Digest a is a Monoid under HashAlgorithm a, I've commited the correct (I hope) instance (orphan) in Crypto.Hash

vincenthz:
I don't think `Monoid` make sense for `Digest`. I think it breaks pretty much all laws for monoid. For starter:

```
mappend x mempty /= x
```

paolino:
I see, there is no mempty for Digest a. I hope Read is not so wrong.
Thanks

vincenthz:
On the other hand, `Read` is clearly useful, but I don't think it's correct. IIRC, it's eating everything until the end of the `String`, then calling digestFromByteString.

It should compute how much data is needed (known length of a `Digest`), then eat it, then unbase16 it to a digest.

paolino:
I've fixed it.
I don't know if the (*2) is intended.
I think I'm hiding a possible implementation error in a Nothing, it should be "The impossible happened...." on the last bind, if digestFromByteString can only fail for short lengths

cryptonite.cabal
```diff
@@ -192,7 +192,7 @@ Library
                    , bytestring
                    , memory >= 0.8
                    , ghc-prim
-  ghc-options:       -Wall -fwarn-tabs -optc-O3 -fno-warn-unused-imports
+  ghc-options:       -Wall -fwarn-tabs -optc-O3 -fno-warn-unused-imports -fobject-code
```
vincenthz, line 195: this shouldn't be here

ixmatus:
I wanted to drop in and say I'd like to see the Read instance since I just implemented something similar by-hand.

centromere:
@paolino Are you still interested in pursuing this pull request?

osa1:
:+1: we also need a `Read` instance.

ocheron:
Closing: `Monoid` part is not needed and `Read` part is replaced by #219.

PR #80, opened 2016-04-18
[Ed448] Fixed incorrect base point

centromere:
The base point was incorrectly specified causing public keys derived by `toPublic` to be invalid. Unit tests were added to check for this going forward.


vincenthz:
thanks, looking definitely less dodgy now with everything escaped. good catch :)

centromere:
Thanks. Do you have any other features slated for 0.16?

vincenthz:
I think just this basepoint-fix is enough to do a release; would like to merge the other PRs, but they can be merged for 0.17 or later.

PR #85, opened 2016-04-27
Example of symmetric encryption in documentation.

yogsototh:
It took me way too long and to search in you tests to discover how to encrypt a file using a key.
So I believe it would be a good idea to just provide some example of usage.

I believe `ctrCombine` with AES256 should be good enough for most usage of symmetric encryption.

So I simply gave an example. I put the example in a `Crypto.Tutorial` as it is done for the `Turtle` library which is praised for its documentation.

### DIFF
--- a/Crypto/Tutorial.hs
+++ b/Crypto/Tutorial.hs
@@ -1,0 +1,34 @@
+{-# OPTIONS_GHC -fno-warn-unused-imports #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+{-| How to use @cryptonite@
+
+> -- | Beware MUST BE 256bits as we use AES256
+> import Data.ByteString (ByteString)
+> import Crypto.Cipher.AES (AES256)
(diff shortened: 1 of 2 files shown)


vincenthz:
I think it's a great idea, but is there a way we could typecheck / execute the tutorial code ? I would like to have it as part of the automated testing to make sure we don't ship tutorial that are useless

yogsototh:
Yes, trying to find a solution I discovered the `doctest` package:

https://github.com/sol/doctest#readme

it could be used simply like this:

https://github.com/Gabriel439/Haskell-Turtle-Library/blob/master/test/Main.hs

So you could provide testing on the documentation.

tekul:
I use doctest in one of my projects and it works pretty well. It can also be added as a separate test suite in the cabal file.

I think it'd be a good idea to use a more realistic AES256 key in the example. Perhaps hex or base64 decode an actual random key of the correct size?

vincenthz:
merged it now to start experimenting with `doctest`, but definitely further doc improvement will be good

PR #92, opened 2016-07-24
Arithmetic over F2m

Bodigrim:
The aim of the pull request is to refactor `Crypto.Number.F2m` module.

My starting point was a comment near `squareF2m` routine, stating that for unknown reason `squareF2m` is slower than `mulF2m`. So I have written a benchmark for functions over F2m, which clearly showed that `squareF2m` is 2.5-3x slower than `mulF2m`. Another interesting fact was that the benchmark failed with out of memory error on `invF2m` routine with 256bit arguments.

This memory leak led me to writing more or less comprehensive test suite for `Crypto.Number.F2m`. The tests showed thal along with the memory leak in `invF2m` there are many corner cases in other functions. They fall into two categories. The first category includes division by zero modulus - definitely, routines should throw an error in this case.

The second category of test failures was caused by negative numbers. The module represents polynomials over F2m (isomorphic to bit vectors) by `Integer` values. For example, `1` represents the polynomial of zero degree `1`. However, it is not clear which polynomial is represented by `-1`: is it `x^63 + x^62 + ... + 1` or `x^31 + x^30 + ... + 1`? That means that the result of computations is platform dependent: basically, encryption over elliptic curve on 64bit architecture cannot be decrypted on 32bit architecture.

More over, with negative `Integers` there are two very different representations for every polynomial. On 64bit machine `x^63 + x^62 + ... + 1` is represented both by `-1` (`S# I#`) and by `18446744073709551615` (`Jp# BigNat`). It does not make much sense to allow both of them.

That said, I decided to restrict input of routines over F2m to non-negative `Integer`. One can check that given non-negative inputs they produce non-negative outputs.

Further, I have fixed the memory leak in `invF2m`. It was caused by infinite recursion. The underlying extended GCD algorithm was rewritten in a more clear and robust way (now it handles correctly all polynomials, not only irreducible ones).

That said, I was able to return to speeding up `squareF2m`. I have rewritten it as a very simple one-liner, improving its running time 3x. Now it is 10% faster than `mulF2m`. This comparison is not very impressive - one may expect `squareF2m` to run in linear time, while `mulF2m` runs in quadratic time. The reason is that `setBit` on `Integer` has linear complexity (not `O(1)` as one may expect).

In future it may be worth of effort to replace `Integer` with true bit vectors (e. g., from `bitvec` package): they would both solve mess with signs and speed up `squareF2m` asymptotically, because `setBit` is a constant-time operation for bit vector.

BTW I have tried to use Karatsuba multiplication for `mulF2m`, but it appeared to be fruitful only for very large polynomials (5000 bits and more), so I reverted the change.

As a detour, I have fixed a bug in `Crypto.PubKey.ECC.Prim.pointMul`, when a point on `CurveF2m` was multiplied by negative factor. The negation on `CurveFP` and `CurveF2m` works differently, see new `negatePoint` routine. It also simplified a bit the code of `pointAdd`.

### DIFF
--- a/Crypto/Number/F2m.hs
+++ b/Crypto/Number/F2m.hs
@@ -9,6 +9,7 @@
 -- not optimal and it doesn't provide protection against timing
 -- attacks. The 'm' parameter is implicitly derived from the irreducible
 -- polynomial where applicable.
+
 module Crypto.Number.F2m
     ( BinaryPolynomial
     , addF2m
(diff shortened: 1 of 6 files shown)


Bodigrim:
@vincenthz any comments?

vincenthz:
really nice thanks ! definitely very thorough and well wrapped (extra bonus for tests and benchs !)

PR #95, opened 2016-07-28
Derive Show instance for CryptoFailable

### LINKED ISSUE #50: Show instance for CryptoFailable is wrong
CryptoFailable has a manually written Show instance that neither surrounds itself or its argument with parentheses correctly. I think that the correct solution is to just derive the correct instance, but for some reason the incorrect one is manually written.

I'll be happy to provide a patch if desired manually implementing the correct instance or deriving the Show instance if that would be helpful; I just need to know which you'd want.

glguy:
Fixes #50

The derived instances use precedences to decide when
parentheses are appropriate.


vincenthz:
that's fine ! thanks.

PR #97, opened 2016-08-25
Fix ECDSA with larger hash size

ocheron:
As reported by @xnyhps in vincenthz/hs-tls#152 the hash output is not truncated correctly in ECDSA signature generation and verification.

This PR fixes the bit shift and adds test cases where truncation occurs. I copied test vectors from RFC 6979 because that is the most convenient I found and gives a good range of curve sizes and hash sizes. I kept only P-\* curves because others are much slower.

Please note that the initial error `HandshakeFailed Error_EOF` still requires another modification to tls.

### DIFF
--- a/Crypto/PubKey/ECC/ECDSA.hs
+++ b/Crypto/PubKey/ECC/ECDSA.hs
@@ -20,12 +20,14 @@ import Crypto.Random.Types
 import Data.Bits (shiftR)
 import Crypto.Internal.ByteArray (ByteArrayAccess)
 import Data.Data
+import Crypto.Number.Basic (numBits)
 import Crypto.Number.ModArithmetic (inverse)
 import Crypto.Number.Serialize
 import Crypto.Number.Generate
 import Crypto.PubKey.ECC.Types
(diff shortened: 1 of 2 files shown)


xnyhps:
There might be other situations where the old truncation method is required, it is after all how it is specified in FIPS 186-3. Not everyone might like changing the semantics of `tHash` like this.

ocheron:
I guess sometimes you must break interoperability with past versions of the library in order to gain interoperability with other implementations.

When you take the leftmost N bits of the hash value, you should take the same bit positions regardless of the bit values 0 or 1.

i.e.
- if the hash output is 0x11223344, the 8 leftmost bits give 0x11
- if the hash output is 0x00112233, the 8 leftmost bits are 0x00, not 0x11

I don't see in the documents where the current implementation is required.

vincenthz:
that seems to me to make more sense than the other method; Although if the other method is really required (I'm not sure it is right now), we can find a way to distinguish the functions

well done for porting a good set of tests too ! thanks

PR #98, opened 2016-09-03
Add Byteable instance.

vaibhavsagar:
This seems like an innocuous change, is there a reason it hasn't already been implemented?


vincenthz:
`Byteable` is not necessary anymore, everything is in `memory`. you can use `Data.ByteArray.convert` to do the equivalent of `toBytes`

vaibhavsagar:
I see, thanks for getting back to me!

PR #101, opened 2016-09-10
Fix ECDH/ECDSA when scalar and coordinate bit sizes differ

ocheron:
I found more bit-alignment issues when testing handshake between `tls` and
OpenSSL using less frequent elliptic curves.  In the implementation of ECDH
and ECDSA there is sometimes a confusion between the curve order `n` and the
size of the finite-field.

The public-key components exchanged during handshake are point coordinates
and should use the bit size of the finite field, instead of the order `ecc_n`
which is for scalars.  For P-256 both sizes match, but for other curves they
may differ.

This pull request changes the function `curveSizeBits` exported from
`Crypto.PubKey.ECC.Types` to return the bit length of the finite field instead
of the order `n`.

i.e. change:

``` haskell
curveSizeBits = numBits . ecc_n . common_curve
```

to the following:

``` haskell
curveSizeBits (CurveFP  c) = numBits (ecc_p  c)
curveSizeBits (CurveF2m c) = numBits (ecc_fx c) - 1
```

I believe having this two-equation definition here is more useful.  And the
change would then make it easier to fix byte lengths computed in `tls` and
`x509-validation` (just use a function from the existing API instead of
duplicating code).

`curveSizeBits` is used in cryptonite module `Crypto.PubKey.ECC.DH`, which
sometimes fails with the current definition based on `ecc_n`.  None of
48 reverse dependencies on hackage reference `curveSizeBits` currently.

I prepared changes to other packages here for preview:
- [x509-validation](https://github.com/vincenthz/hs-certificate/compare/master...ocheron:ecc-bitsize-fixes)
- [tls](https://github.com/vincenthz/hs-tls/compare/master...ocheron:ecc-bitsize-fixes)

### DIFF
--- a/Crypto/PubKey/ECC/Types.hs
+++ b/Crypto/PubKey/ECC/Types.hs
@@ -164,7 +164,8 @@ curvesOIDs =
 
 -- | get the size of the curve in bits
 curveSizeBits :: Curve -> Int
-curveSizeBits = numBits . ecc_n . common_curve
+curveSizeBits (CurveFP  c) = numBits (ecc_p  c)
+curveSizeBits (CurveF2m c) = numBits (ecc_fx c) - 1
 
 -- | Get the curve definition associated with a recommended known curve name.
 getCurveByName :: CurveName -> Curve


ocheron:
Comparison of `curveSizeBits` results:

| Curve | Old | New | Differ? |
| --- | --- | --- | --- |
| SEC_p112r1 | 112 | 112 |  |
| SEC_p112r2 | 110 | 112 | Y |
| SEC_p128r1 | 128 | 128 |  |
| SEC_p128r2 | 126 | 128 | Y |
| SEC_p160k1 | 161 | 160 | Y |
| SEC_p160r1 | 161 | 160 | Y |
| SEC_p160r2 | 161 | 160 | Y |
| SEC_p192k1 | 192 | 192 |  |
| SEC_p192r1 | 192 | 192 |  |
| SEC_p224k1 | 225 | 224 | Y |
| SEC_p224r1 | 224 | 224 |  |
| SEC_p256k1 | 256 | 256 |  |
| SEC_p256r1 | 256 | 256 |  |
| SEC_p384r1 | 384 | 384 |  |
| SEC_p521r1 | 521 | 521 |  |
| SEC_t113r1 | 113 | 113 |  |
| SEC_t113r2 | 113 | 113 |  |
| SEC_t131r1 | 131 | 131 |  |
| SEC_t131r2 | 131 | 131 |  |
| SEC_t163k1 | 163 | 163 |  |
| SEC_t163r1 | 162 | 163 | Y |
| SEC_t163r2 | 163 | 163 |  |
| SEC_t193r1 | 193 | 193 |  |
| SEC_t193r2 | 193 | 193 |  |
| SEC_t233k1 | 232 | 233 | Y |
| SEC_t233r1 | 233 | 233 |  |
| SEC_t239k1 | 238 | 239 | Y |
| SEC_t283k1 | 281 | 283 | Y |
| SEC_t283r1 | 282 | 283 | Y |
| SEC_t409k1 | 407 | 409 | Y |
| SEC_t409r1 | 409 | 409 |  |
| SEC_t571k1 | 570 | 571 | Y |
| SEC_t571r1 | 570 | 571 | Y |

Impact examples:
- SEC_p160r1: 21 bytes expected instead of 20
- SEC_t233k1: 29 bytes expected instead of 30

Handshake errors observed because of this:
- Error_Protocol ("bad SignatureECDSA for ecdhparams",True,HandshakeFailure)
- Error_Misc "pointSize\nCallStack (from HasCallStack):\n  error, called at ./Network/TLS/Crypto/ECDH.hs:76:15 in tls-1.3.8-BicJFAc6lQ7HHri6idxYEy:Network.TLS.Crypto.ECDH"
- 140735210491984:error:1408F119:SSL routines:SSL3_GET_RECORD:decryption failed or bad record mac:s3_pkt.c:532: (remote side)
- i2ospOf_: integer is larger than expected

vincenthz:
Thanks, that looks much better. And well done for the thoroughness

PR #103, opened 2016-10-09
Implement the XSalsa20 stream cipher

brandonhamilton:


vincenthz:
looking good ! thanks !

PR #104, opened 2016-10-09
Correct XSalsa20 initialization when using 8/12 rounds

brandonhamilton:
Sorry, previous commit had a minor error


vincenthz:
Thanks !

PR #105, opened 2016-10-12
Parameterized EC and ECDH

kazu-yamamoto:
This patch extends the ecc branch discussed in https://github.com/vincenthz/hs-tls/issues/150


vincenthz:
Why did you remove ECIES, and change ECC.DH ?

kazu-yamamoto:
No reason. I just tried. If you want to keep ECIES and ECC.DH, I will extend ECIES. What I need is your direction!

kazu-yamamoto:
Closing this due to #111.

PR #110, opened 2016-11-06
Fix unaligned memory accesses; fixes #108

### LINKED ISSUE #108: workaround for alignment issue
```
Author: Steve Langasek <steve.langasek@ubuntu.com>
Description: fix alignment of memory blocks used by SHA3
 SHA3 works in 64-bit chunks, but the incoming data pointer can be at any
 address.  Copy our data to an aligned address, to avoid SIGBUS on certain
 platforms.
 .
 This is not the only alignment issue in the code, but it is the one that
 manifests as SIGBUS on the most architectures.

Index: haskell-cryptonite-0.20/cbits/cryptonite_sha3.c
===================================================================
--- haskell-cryptonite-0.20.orig/cbits/cryptonite_sha3.c
+++ haskell-cryptonite-0.20/cbits/cryptonite_sha3.c
@@ -23,6 +23,7 @@
  */

 #include <stdint.h>
+#include <stdlib.h>
 #include <string.h>
 #include "cryptonite_bitfn.h"
 #include "cryptonite_sha3.h"
@@ -107,6 +108,7 @@ void cryptonite_sha3_init(struct sha3_ct
 void cryptonite_sha3_update(struct sha3_ctx *ctx, const uint8_t *data, uint32_t len)
 {
    uint32_t to_fill;
+   uint64_t *data_aligned = NULL;

    to_fill = ctx->bufsz - ctx->bufindex;

@@ -124,6 +126,13 @@ void cryptonite_sha3_update(struct sha3_
            ctx->bufindex = 0;
    }

+   /* fix up alignment if necessary */
+   if (len && (unsigned long) data & 7) {
+           data_aligned = malloc(len);
+           memcpy(data_aligned, data, len);
+           data = (uint8_t *) data_aligned;
+   }
+
    /* process as much ctx->bufsz-block */
    for (; len >= ctx->bufsz; len -= ctx->bufsz, data += ctx->bufsz)
            sha3_do_chunk(ctx->state, (uint64_t *) data, ctx->bufsz / 8);
@@ -133,6 +142,7 @@ void cryptonite_sha3_update(struct sha3_
            memcpy(ctx->buf + ctx->bufindex, data, len);
            ctx->bufindex += len;
    }
+   free(data_aligned);
 }

 void cryptonite_sha3_finalize(struct sha3_ctx *ctx, uint32_t hashlen, uint8_t *out)
```

jrtc27:


vincenthz:
valiant effort, but this is the very serious side effect of slowing down aligned data processing massively (and architecture that doesn't suffer from some unaligned data).

jrtc27:
Do you have tests to back that up? The functions are static inline, and any worthwhile compiler _should_ be able to optimise that. Alternatively I can turn them into memcpy's with byte swaps as needed?

glaubitz:
@vincenthz The problem is that the unaligned access will trigger bus errors on some architectures like sparc64 and therefore result in a crash.

vincenthz:
@jrtc27 I have done lots of careful optimisation for the hash functions, and I'm pretty sure that byte access kill performance massively. As to the compiler optimisation, there's no chance the compiler is able to turn byte memory access that might be unaligned, into faster word access because a function is static inline. memcpy is not faster to load word into registers.

@glaubitz I'm well aware of the problem with unalignment. Since the number of people owning a sparc64 (and also running haskell) is much much smaller than people owning x86-64 cpus, I won't  de-optimise those functions for 99.9% of the people.

vincenthz:
Also some data point to distinguish in alignment issues:
- data that is allocated by haskell and should be already aligned correctly; for example the digest and context memory should be allocated to 8 bytes by `memory`, and thus shouldn't trigger unalignment issue.
- other data either on the stack that is not aligned, but could be.
- user data (generally unaligned).

jrtc27:
@vincenthz As it stands, the hash functions include undefined behaviour, including on x86. It is not guaranteed to do the right thing on x86, but it's effectively guaranteed (although not necessarily; some SSE instructions require stricter alignment). Are you telling me you value performance over correctness? If you really care that much about x86 performance, I can put `#if !defined(__i386__) && !defined(__x86_64__)` around the "slower" versions and use the old versions for x86? But, rather than constantly bringing up issues with this code, could you please instead offer _constructive_ criticism on how to make it acceptable to you?

jrtc27:
I have also since discovered that the AES implementation is plagued with casts to `block128 *` that the test suite didn't even catch; it took another pure-Haskell library's test suite attempting to use it to discover this. This is a very real problem.

vincenthz:
none of the hash functions use SSE instructions. related to block128, there's nothing much different to what the hash functions already do.

What's acceptable:
- no portability tax in performance. >byte word access need to remain >byte word access with aligned data and architecture that are mostly oblivious (performance and correctness wise) to alignment issue on <= 8 bytes.
- minimum portability tax in source: need an abstraction that hide the architectural difference as much as possible so the code is not #ifdef everywhere.
- Make sure that guarantee aligned data by the haskell layer doesn't add any unaligned handling for no reason.

vincenthz:
some modules are now a bit more friendly towards alignments guarantees/requirements, but it's not complete.

PR #111, opened 2016-11-08
Extending ECIES

kazu-yamamoto:
According to the discussion on #105, I only extend ECIES based on the ecc branch in #40.
This implementation is actually tested in my TLS 1.3 (or would be called TLS 2.0) extension to the tls library.


kazu-yamamoto:
When I try to support X25519 in ECIES, I understand the intention of your original code.
I would like to withdraw this PR and send a new PR when ready.

PR #112, opened 2016-11-16
Extending ECC and ECIES

kazu-yamamoto:
This patches extend the rebased `ecc` branch to support ECDH, P384 and X25521.
I confirmed that this can be use in the `tls` library.

This also extends PRK which is necessary to implement TLS1.3.

Please give me comments.

### DIFF
--- a/Crypto/ECC.hs
+++ b/Crypto/ECC.hs
@@ -1,0 +1,112 @@
+-- |
+-- Module      : Crypto.ECC
+-- License     : BSD-style
+-- Maintainer  : Vincent Hanquez <vincent@snarc.org>
+-- Stability   : experimental
+-- Portability : unknown
+--
+-- Elliptic Curve Cryptography
+--
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TypeFamilies #-}
+module Crypto.ECC
+    ( Curve_P256R1(..)
+    , Curve_P521R1(..)
+    , EllipticCurve(..)
+    , EllipticCurveDH(..)
+    , EllipticCurveArith(..)
+    , KeyPair(..)
+    , SharedSecret(..)
+    ) where
+
+import qualified Crypto.PubKey.ECC.P256 as P256
+import qualified Crypto.PubKey.ECC.Types as H
+import qualified Crypto.PubKey.ECC.Prim as H
+import           Crypto.Random
+import           Crypto.Internal.ByteArray (ByteArrayAccess, ScrubbedBytes)
+import           Data.Function (on)
+
+-- | An elliptic curve key pair composed of the private part (a scalar), and
+-- the associated point.
+data KeyPair curve = KeyPair
+    { keypairGetPublic  :: !(Point curve)
+    , keypairGetPrivate :: !(Scalar curve)
+    }
+
+newtype SharedSecret = SharedSecret ScrubbedBytes
+    deriving (Eq, ByteArrayAccess)
+
+class EllipticCurve curve where
+    -- | Point on an Elliptic Curve
+    data Point curve  :: *
+
+    -- | Scalar in the Elliptic Curve domain
+    data Scalar curve :: *
+
+    -- | get the order of the Curve
+    curveGetOrder :: curve -> Integer
+
+    -- | get the curve related to a point on a curve
+    curveOfPoint :: Point curve -> curve
+
+    -- | get the curve related to a curve's scalar
+    curveOfScalar :: Scalar curve -> curve
+
+    -- | get the base point of the Curve
+    curveGetBasePoint :: Point curve
+
+    -- | Generate a new random scalar on the curve.
+    -- The scalar will represent a number between 1 and the order of the curve non included
+    curveGenerateScalar :: MonadRandom randomly => randomly (Scalar curve)
+
+    -- | Generate a new random keypair
+    curveGenerateKeyPair :: MonadRandom randomly => randomly (KeyPair curve)
+
+class EllipticCurve curve => EllipticCurveDH curve where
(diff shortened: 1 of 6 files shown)


vincenthz (commented):
I don't think that's the right thing to do. if you really need the PRK in binary format, you should make it part of ByteArrayAccess / ByteArray class

Crypto/ECC.hs
```diff
@@ -0,0 +1,236 @@
...
+    -- The scalar will represent a number between 1 and the order of the curve non included
+    curveGenerateScalar :: MonadRandom randomly => randomly (Scalar curve)
+
+    -- | Generate a new random keypair
+    curveGenerateKeyPair :: MonadRandom randomly => randomly (KeyPair curve)
+
+    encodePoint :: Point curve -> ByteString
+    decodePoint :: ByteString -> Point curve
+
+instance {-# OVERLAPPABLE #-} Show (Point a) where
```
vincenthz, line 76: That seems like the ugly conclusion of having the associated type as newtype instead of types; really shouldn't be here

Crypto/ECC.hs
```diff
@@ -0,0 +1,236 @@
...
+encodeECPoint x y siz = B.concat [uncompressed,xb,yb]
+  where
+    uncompressed = B.singleton 4
+    xb = i2ospOf_ siz x
+    yb = i2ospOf_ siz y
+
+decodeECPoint :: ByteString -> (Integer,Integer)
+decodeECPoint mxy = (x,y)
+  where
+    xy = B.drop 1 mxy -- dropping 4 (uncompressed)
```
vincenthz: it really really shouldn't be a drop 1. at the very least it should be a case that check for 4, and put a note for the 2,3 format. Also this shouldn't use `ByteString` but should do things through the ByteArray / ByteArrayClass

Crypto/KDF/HKDF.hs
```diff
@@ -15,18 +15,32 @@ module Crypto.KDF.HKDF
...
 import           Crypto.MAC.HMAC
 import           Crypto.Internal.ByteArray (ScrubbedBytes, Bytes, ByteArray, ByteArrayAccess)
 import qualified Crypto.Internal.ByteArray as B
+import qualified Data.ByteString as BS
 
 -- | Pseudo Random Key
 data PRK a = PRK (HMAC a) | PRK_NoExpand ScrubbedBytes
     deriving (Eq)
 
+instance Show (PRK a) where
```
vincenthz: I don't think this should have a Show instance for security reason

Crypto/KDF/HKDF.hs
```diff
@@ -15,18 +15,32 @@ module Crypto.KDF.HKDF
...
+    show (PRK hm) = show (hmacGetDigest hm)
+    show (PRK_NoExpand sb) = show sb
+
+toByteString :: PRK a -> BS.ByteString
```
vincenthz: Should be ByteArray / ByteArrayAccess not direct use of ByteString

Crypto/PubKey/Curve25519.hs
```diff
@@ -110,3 +119,22 @@ foreign import ccall "cryptonite_curve25519_donna"
                            -> Ptr Word8 -- ^ secret
                            -> Ptr Word8 -- ^ basepoint
                            -> IO ()
+
+generateSecretKey :: MonadRandom m => m SecretKey
+generateSecretKey = return $ unsafeDoIO $ do
```
vincenthz, line 124: you really don't want to put things into a bytestring and then moved it to a Scrubbed bytes. it does defeat the purpose. getRandomBytes should gives you Scrubbed Bytes too if you want.

Crypto/KDF/HKDF.hs
kazu-yamamoto: Done.

Crypto/ECC.hs
kazu-yamamoto: I have no idea on how to make `Point` an instance of `ByteArrayAccess`.
Would you show me code?

Crypto/KDF/HKDF.hs
kazu-yamamoto: Done.

Crypto/PubKey/Curve25519.hs
kazu-yamamoto: Done.

kazu-yamamoto:
Now `PRK` is an instance of `ByteArrayAccess`.

Crypto/ECC.hs
kazu-yamamoto: Is this comment for `instance` rather than `encodePoint`?

Crypto/ECC.hs
kazu-yamamoto: Done.

Crypto/ECC.hs
kazu-yamamoto: `Point` and `Scalar` should be instance of `Eq` and `Show` because upper layer data structures in `tls` require so.

kazu-yamamoto: I have no idea on how to make `Point` and `Scalar` instances of `Eq` and `Show` except this approach.

kazu-yamamoto:
All issues except one, which I cannot solve, are fixed.
Please merge this PR and fix the last issue by yourself.

kazu-yamamoto:
Note that those patches are tested with the `tls13` branch of my fork of `tls`.
WAI applications can communicate with Firefox Nightly, Chrome Canary and picotls in TLS 1.3 based on `cryptonite` with those patches.
P256/P384/P512/X25519 are tested.

vincenthz:
now tracked by #114

PR #114, opened 2016-12-01
Rewriting ECC

vincenthz:


Crypto/ECC/Simple/Prim.hs
```diff
@@ -0,0 +1,197 @@
...
+pointMul (Scalar n) p
+    | n == 0    = PointO
+    | n == 1    = p
+    | odd n     = pointAdd p (pointMul (Scalar (n - 1)) p)
+    | otherwise = pointMul (Scalar (n `div` 2)) (pointDouble p)
+
+-- | Elliptic curve double-scalar multiplication (uses Shamir's trick).
+--
+-- > pointAddTwoMuls c n1 p1 n2 p2 == pointAdd c (pointMul c n1 p1)
+-- >                                             (pointMul c n2 p2)
```
ocheron: In the equation you can remove the argument `c`.

vincenthz: Thanks for pointing it out. done

vincenthz:
@kazu-yamamoto head's up I've changed the API significantly to address some issues. that very likely break any code depending on it, and it may have introduced bugs during the rewrite. let me know if there's anything missing

PR #116, opened 2016-12-05
Fix ecc2

kazu-yamamoto:
I found two bugs in P256.

X25519 works well. I have not tested P384 and P521 yet.


vincenthz:
Thanks. sorry about introducing the bugs

kazu-yamamoto:
No problem.
Thank you for cleaning up my code and the merges!

Issue #113, opened 2016-11-22
Any valid AES key size works for AES256?


tmciver:
I'm initializing an `AES256` `Cipher` as seen here: https://github.com/unisonweb/unison/blob/5b22506804c844bc09c05673061d93c8b6d8281f/node/src/Unison/Runtime/Cryptography.hs#L62 but I found that there is no problem with using a 128-bit key.  Is this expected behavior?

I see that the key length for `AES256` is fixed here: https://github.com/haskell-crypto/cryptonite/blob/47cb6ebdeaa6149b9cd2e70422e037ace3cb33ae/Crypto/Cipher/AES.hs#L45 but it is apparently not checked.

vincenthz:
Looks like a small bug where it will give you a AES128 context wrapped in `AES256` because of the 128 bit key. care to make a fix ?

tmciver:
Sure, I'll take a shot at it.  A question though: do the `AES128`, `AES192` and `AES256` types represent "keys"?  You mentioned "context" (as do the comments); is that something different than just a key?

vincenthz:
yes, the first step of AES is preparing a key (128, 192, 256 bits) into a many rounds key (12 to 14 128 bits keys) which is what is called context here.

PR #118, opened 2016-12-07
Add key length validation to several AES Ciphers.

tmciver:
Addresses #113 

Please give feedback if this work is not in keeping with the style of the cryptonite codebase.  I'm particularly interested in any thoughts concerning the proper location of the function `validateKeySize` since it is rather generic and should work with any `Cipher` so probably does not belong in `Crypto.Cipher.AES`.

### DIFF
--- a/Crypto/Cipher/AES.hs
+++ b/Crypto/Cipher/AES.hs
@@ -18,6 +18,8 @@ import Crypto.Cipher.Types.Block
 import Crypto.Cipher.AES.Primitive
 import Crypto.Internal.Imports
 
+import Data.ByteArray as BA
+
 -- | AES with 128 bit key
 newtype AES128 = AES128 AES
     deriving (NFData)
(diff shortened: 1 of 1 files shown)


vincenthz:
Overall fine, but it might be simpler to just assume the key size per cipher (just like the above line) instead of jumping through an undefined and a generic function

vincenthz:
Just going to merge it to make a new release, but I'll appreciate if you could address the comment in the future (no obligation, the code works, Thanks !)

tmciver:
@vincenthz Cool.  Can you clarify what you'd like to have changed?  Do you mean perhaps having a function that takes the key size as a parameter and just hard-code that at the call site?

vincenthz:
yes exactly a function taking the size directly. it's hardcoded above, so there's not too much problem here, ideally something to avoid, but here it's pretty harmless

Issue #117, opened 2016-12-06
Incorrect authentication tag using ChaChaPoly1305 and chunked lazy bytestrings


crossleydominic:
We seem to be seeing incorrect authentication tag generation when using the ChaChaPoly1305 algorithm with lazy bytestrings.

When reading a lazy bytestring from the network we apply the ChaChaPoly1305 incrementally to the bytestring in chunks (so as not to read the entire bytestring into memory at once). If these chunks are non uniform in size then the resulting authentication tag is not correct. The generated tag seems to be dependant upon the number of chunks and the size of each chunk rather than just the entirety of the binary data in the lazy byte string.

A minimum reproducible example is here:
https://gist.github.com/flashcurd/6f4fd28cbe205afff47e13af2cc1e925

The example encrypts using three different methods. 
1. A strict bytestring is used
2. A lazy bytestring is used with large uniform chunk sizes.
3. A lazy bytestring is used with randomly sized chunks (emulating data read from the network).

Scenarios 1 and 2 generate the same authentication tag. Scenario 3 generates a different authentication tag for each execution of the example program. The key, nonce and file binary are all hardcoded.

The example program MD5s the resulting ciphertext to compare between scenarios. It is the same in all cases. It just seems to be the authentication tag which is affected.

Example output (apologies for the raw dumping of the binary data, but it's good enough to show the differences):

```
"Strict:"
"Strict tag: \"\\ACK\\206\\202=M2\\142|\\147/>\\GS3\\254E\\228\""
"Strict cipher md5: \"\\247s\\177\\140\\\"\\193[|\\154\\233N\\252\\168\\253\\183\\177\""
"---"
"Lazy, same chunk size"
"Number of chunks: 8"
"Chunk sizes: [32752,32752,32752,32752,32752,32752,32752,3172]"
"Lazy tag: \"\\ACK\\206\\202=M2\\142|\\147/>\\GS3\\254E\\228\""
"Lazy cipher md5: \"\\247s\\177\\140\\\"\\193[|\\154\\233N\\252\\168\\253\\183\\177\""
"---"
"Chunked, random chunk size"
"Number of chunks: 233"
"Chunk sizes: [101,25,1749,1907,1836,54,1597...snip...,477,1283,1361]"
"Chunked tag: \"\\147?5\\228\\181\\173\\177|\\207|\\218\\201\\133\\244{\\207\""
[...]

PR #119, opened 2016-12-07
Update the index into a partial Poly1305 buffer instead of overwriting

crossleydominic:
Possible for for #117


vincenthz:
Thanks. I think that's indeed the right fix

PR #123, opened 2016-12-11
Added a couple of alternate key length validation functions for review.

tmciver:
@vincenthz This PR is to address your concerns with PR #118.  I added a couple of alternate versions of a key length validation function.  Let me know if either is what you had in mind.

The main problem I see with using either of these functions is that they ignore they value specified for `KeySizeSpecifier` in the `Cipher` implementation, so the implementor has to specify it twice. And of course there's the problem that those two values could be different!


vincenthz:
It's not quite what I had in mind. I'm thinking a very simple helper function

```haskell
something :: Int -> bytearray -> Key
```

which would be called explicitely with the expected size:

```haskell
instance AES128 ...
   cipherInit = something 16 input
instance AES256
   cipherInit = something 32 input
```

centromere:
@tmciver Are you still interested in pursuing this pull request?

tmciver:
@centromere Sure.  @vincenthz do you still want this change?  I must admit I'm not sure why this would be preferred over the fix in #118.  The change you're suggesting ignores the `KeyLengthSpecifier` declared in the `Cipher` and duplicates it in the function.  What's the purpose of specifying a key size in the `Cipher` if it's not going to be used in the code that verifies the key size?

tmciver:
@vincenthz This isn't exactly what you asked for but at least it gets us away from doing `undefined :: AES128`, etc.  If you don't like it, I'll do it your way.  But I like that this method doesn't require respecifying the key size.  (Full disclosure: I had help with this from someone in #haskell IRC).

vincenthz:
archiving repository

Issue #124, opened 2016-12-22
Generated primes are of incorrect size


colatkinson:
Sometimes when generating primes, they are larger than the selected bit size.
For example, when generating a prime of 5 bits, the output will be 37 (6 bits), 67 for 6 bits, or 257 for 8 bits.

tekul:
Hi, could you give an example of the code you are using please?

colatkinson:
Using current master:

```
Prelude Crypto.Number.Prime> generatePrime 5
29
Prelude Crypto.Number.Prime> generatePrime 5
37
Prelude Crypto.Number.Prime> generatePrime 8
257
Prelude Crypto.Number.Prime> generateSafePrime 5
227
```

The problem persists in compiled code.

Based on looking at the code, I think that this is probably caused by the fact that `findPrimeFrom` and `findPrimeFromWith` do not perform any checks on the size of their results. A simple solution would be to just perform a check in `generatePrime` and `generateSafePrime` to check if the result is less than 2^bits, although I'm not sure if a better method exists.

tekul:
My guess would be that for the expected crypto-related use cases, the function would only be used to generate large primes (to generate an RSA key, for example). For the large number of bits that would be used, the chance of overflowing by an extra bit is small and doesn't have a detrimental effect in any case. @vincenthz can probably clarify.

colatkinson:
It does have a detrimental effect in that if a piece of code is generating an n-bit key, runs code to generate an n-bit key, and gets an (n+1)-bit key, the library is introducing room for unforeseen bugs. If I want to e.g. send a public key with a strict protocol and it doesn't fit into the allotted space, then my application is broken, or at a minimum has to relax some guarantees.

Given that a fix is basically just an if statement, I'm working on a PR.

vincenthz:
yes @tekul 's explanation is the right one, no effort is made to generate a key of a specific bit size, but of at *least* a specific bit size. I don't think generating a number of 5 to 8 bits is what this function was designed for.

One obvious thing, is that we clearly lack documentation here

ocheron:
@vincenthz I have a related question (somewhat): while looking at generation of DHParams I noticed that the slow operation is `generateSafePrime`. The current implementation in `findPrimeFromWith` does not call GMP like `findPrimeFrom` does. Instead it tests every odd integer successively.

Is this on purpose?

[...]

PR #125, opened 2016-12-25
Fix generated primes being too large

colatkinson:
Simple fix for #124, simply by testing if the prime is too big and recurring if so.

Also changes prime generation to search from [2^(n-1), 2^n) instead of [2^(n-1) + 2^(n-2), 2^n] by setting only the top bit, as is the general standard (see e.g. [OpenSSL](https://github.com/openssl/openssl/blob/master/crypto/bn/bn_prime.c#L101)).

### DIFF
--- a/Crypto/Number/Prime.hs
+++ b/Crypto/Number/Prime.hs
@@ -27,6 +27,7 @@ import Crypto.Number.Basic (sqrti, gcde)
 import Crypto.Number.ModArithmetic (expSafe)
 import Crypto.Random.Types
 import Crypto.Random.Probabilistic
+import Crypto.Error
 
 import Data.Bits
 
@@ -40,22 +41,44 @@ isProbablyPrime !n
     | primalityTestFermat 50 (n`div`2) n                   = primalityTestMillerRabin 30 n
     | otherwise                                            = False
 
--- | generate a prime number of the required bitsize
+-- | generate a prime number of the required bitsize (i.e. in the range
+--   [2^(b-1)+2^(b-2), 2^b)).
+--
+--   May throw a CryptoError_PrimeSizeInvalid if the requested size is less
+--   than 5 bits, as the smallest prime meeting these conditions is 29.
+--   This function requires that the two highest bits are set, so that when
+--   multiplied with another prime to create a key, it is guaranteed to be of
+--   the proper size.
 generatePrime :: MonadRandom m => Int -> m Integer
 generatePrime bits = do
-    sp <- generateParams bits (Just SetTwoHighest) True
-    return $ findPrimeFrom sp
+    if bits < 5 then
+        throwCryptoError $ CryptoFailed $ CryptoError_PrimeSizeInvalid
+    else do
+        sp <- generateParams bits (Just SetTwoHighest) True
+        let prime = findPrimeFrom sp
+        if prime < 1 `shiftL` bits then
+            return $ prime
+        else generatePrime bits
 
 -- | generate a prime number of the form 2p+1 where p is also prime.
 -- it is also knowed as a Sophie Germaine prime or safe prime.
 --
 -- The number of safe prime is significantly smaller to the number of prime,
 -- as such it shouldn't be used if this number is supposed to be kept safe.
+--
+-- May throw a CryptoError_PrimeSizeInvalid if the requested size is less than
+-- 6 bits, as the smallest safe prime with the two highest bits set is 59.
 generateSafePrime :: MonadRandom m => Int -> m Integer
 generateSafePrime bits = do
-    sp <- generateParams bits (Just SetTwoHighest) True
-    let p = findPrimeFromWith (\i -> isProbablyPrime (2*i+1)) (sp `div` 2)
-    return (2*p+1)
+    if bits < 6 then
+        throwCryptoError $ CryptoFailed $ CryptoError_PrimeSizeInvalid
+    else do
+        sp <- generateParams bits (Just SetTwoHighest) True
+        let p = findPrimeFromWith (\i -> isProbablyPrime (2*i+1)) (sp `div` 2)
+        let val = 2 * p + 1
+        if val < 1 `shiftL` bits then
+            return $ val
+        else generateSafePrime bits
 
 -- | find a prime from a starting point where the property hold.
 findPrimeFromWith :: (Integer -> Bool) -> Integer -> Integer
--- a/Crypto/Number/Prime.hs
+++ b/Crypto/Number/Prime.hs
@@ -38,7 +38,8 @@ import Data.Bits
 isProbablyPrime :: Integer -> Bool
 isProbablyPrime !n
     | any (\p -> p `divides` n) (filter (< n) firstPrimes) = False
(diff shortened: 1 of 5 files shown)


vincenthz:
I don't have in hand the reference, but I'm pretty sure I've seen some "official" RSA document where the number generation is done with the 2 highest bit set. I want to double check before changing the range of the function. It would really be useful to improve the documentation here and make sure we have explicit reference to documents and what other libraries have done here before more actions.

Also It's very likely that the function will spin forever for a small bit size, as there wouldn't be any prime size in a small range that qualify. While the retrying behavior could be happening easily for smaller size, I think the probability for number that this function typically should be used for (>= 1024 bits) is really really small, that is very very unlikely to happen.

colatkinson:
Looking at the [NIST FIPS 186-4](http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf) section A.1.1.1, the only bits mentioned are the MSB and LSB. The former is to ensure that e.g. 2 is not generated as a 2048-bit prime, and the latter just because primes > 2 have to be odd.

And I agree that the size I was testing on is far from typical. A library I'm using used primes of that small size in example code (understanding what crypto functions are doing is much easier on trivially small examples), which is how I ran across the bug. While the probability of a prime of incorrect size being generated is low, it is nonzero, and the fix is simple enough.

As for the issue of the code hanging if too-small a size is specified, that can probably be handled trivially by just throwing an error if e.g. bits < 2. Is this a good solution in your view?

vincenthz:
Yes, I don't think the API should change in this case, which leave us with an async error. For invalid use that is fine, provided it's correctly documented above. Although I prefer named error from `Crypto.Error` and thrown than calling error with a random string.

vincenthz:
Related to the 2 highest bits, I'm going to try to hunt this down, but I think it's related to an elusive RSA key generation document that is not public for some reason/need to pay to access. really unfortunate situation.

ocheron:
There is an explanation related to multiplication. If n-bit primes p and q have two high bits set, product pq has 2n bits.

http://crypto.stackexchange.com/questions/39779/determining-intervals-containing-the-smallest-prime-p-of-an-rsa-modulus

colatkinson:
@ocheron Gotcha, that makes sense. I've reverted the change.

@vincenthz I've added some error handling and corresponding documentation. While attempting to find the smallest acceptable number of bits, I encountered a bug in the implementation of `isProbablyPrime` wherein any number lower than 100 would return a false negative on `primalityTestFermat`. The starting point is set as half of the number, and the hard coded 50 iterations would exceed the number being tested.

I fixed this by setting the iteration as the minimum of 50 or n/2, and added documentation for this behavior.

ocheron:
IMHO we can keep the spinning because there is not much care taken about the
result bitsize in calling code like `Crypto.PubKey.RSA`. This will also force
evaluation of the result (currently `generateSafePrime` returns an expensive thunk).

In `tests/Number.hs` the property "generate-prime" deals with this exact issue, so that would become useless now:
https://github.com/haskell-crypto/cryptonite/blob/5f72788041b58c82e6b25c0c1e1294982eff7ed9/tests/Number.hs#L47

Crypto/Number/Prime.hs
```diff
@@ -37,25 +38,54 @@ import Data.Bits
 isProbablyPrime :: Integer -> Bool
 isProbablyPrime !n
     | any (\p -> p `divides` n) (filter (< n) firstPrimes) = False
-    | primalityTestFermat 50 (n`div`2) n                   = primalityTestMillerRabin 30 n
+    | primalityTestFermat iters halfn n                    = primalityTestMillerRabin 30 n
     | otherwise                                            = False
+    where
+        halfn = n `div` 2
+        iters = fromInteger (min 50 halfn)
```
ocheron: After testing small primes, a short path returning `True` when `n >= 2 && n <= 2903` would avoid calling `primalityTestFermat` completely.

Crypto/Number/Prime.hs
```diff
@@ -37,25 +38,54 @@ import Data.Bits
...
+    if bits < 5 then
+        throwCryptoError $ CryptoFailed $ CryptoError_PrimeSizeInvalid
+    else do
+        sp <- generateParams bits (Just SetTwoHighest) True
+        let prime = findPrimeFrom sp
+        if prime < 2^bits then
+            return $ prime
+        else do
+            ret <- generatePrime bits
+            return $ ret
```
ocheron: Using `numBits` from `Crypto.Number.Basic` to test the bit size would probably be more efficient (with GMP at least).

Also
```haskell
else do ret <- generatePrime bits
        return $ ret
```
can be shorten to:
```haskell
else generatePrime bits
```

colatkinson:
If forcing the evaluation of the functions is desirable, could it make sense to change the API so that they return a `Maybe` or an `Either`? Code relying on the functions would need to be slightly changed, but it's a cleaner way of handling this than just allowing an infinite loop.

As for the points made in the review, I'll implement the necessary changes, as they definitely simplify the code.

ocheron:
> could it make sense to change the API so that they return a Maybe or an Either ?

Yes and no... from a user point of view I just expect something like `generatePrime` to actually generate the number. You're supposed to have taken care of impossible cases with `CryptoError_PrimeSizeInvalid`. I don't see why this version would loop forever, provided the DRG does its job correctly (if not, then it's probably not adequate for cryptography).

If you still have doubts and to help convincing others, I would recommend updating the tests in `tests/Number.hs`:
- change "generate-prime" to test the range `bits` ∈ [5..804] instead of [48..847], and remove the case `(bits + 1) == numBits prime` not needed anymore
- add a test for `generateSafePrime` (with a much lower maximum bitsize so it's not too slow)

colatkinson:
Yeah the error handles any cases that would cause an infinite loop. The cause was the function searching for a prime when too few bits were available, and thus there were no fitting numbers. So if that's the main concern, it's already handled.

Also in testing the size, my limited benchmarking suggests that the fastest approach for determining bit size is a bit shift, instead of `numBits` or raising to a power, and so I've implemented that.

However I'll update the tests, as they should definitely fit the current implementation.

ocheron:
Yes I noticed `shiftL`, I think it is OK too.

To keep the history more compact and readable, it would be better if you could squash your commits into logical groups, i.e.
1. anything related to generate(Safe)Prime (b23bd89, fd8d889, 4d997f6, 19e6c28, ba9c4d0, c306203, fc33791)
2. bug fix for isProbablyPrime (7d5b003, 6eb48f3)
3. test suite changes

If I do it myself that will remove your PGP signature.

Also I'm not sure about adding `-fobject-code` because I don't see this option in any other project.

colatkinson:
Okay I'll squash the history. I added `-fobject-code` to make the code compile for GHCI, but forgot to remove it at the time, and so it serves no real purpose. I'll remove it.

ocheron:
Thanks, with the tests it looks good.
I'll wait for a second opinion but I think we can merge this.

vincenthz:
Thanks ! Particularly nice to have better documentation

PR #126, opened 2016-12-29
Timing safe equality operator for Digest

dahlia:
Becuase `(==)` operator of `Bytes` is not constant time and `Eq` for `Digest` had been just derived, comparing two `Digest` values could theoretically reveal information for timing analysis through content-based short circuit behavior.

[`Data.ByteArray.constEq`][1] function is a constant time equality test, so it's appropriate for cryptography.  This patch makes `Digest` to prevent timing analysis by implementing `Eq` instance — `(==)` operator — using [`constEq`][1] function instead of deriving `Eq` typeclass.

[1]: http://hackage.haskell.org/package/memory-0.14/docs/Data-ByteArray.html#v:constEq

### DIFF
--- a/Crypto/Hash/Types.hs
+++ b/Crypto/Hash/Types.hs
@@ -51,7 +51,10 @@ newtype Context a = Context Bytes
 
 -- | Represent a digest for a given hash algorithm.
 newtype Digest a = Digest Bytes
-    deriving (Eq,Ord,ByteArrayAccess,NFData)
+    deriving (Ord,ByteArrayAccess,NFData)
+
+instance Eq (Digest a) where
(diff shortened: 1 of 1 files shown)


vincenthz:
you shouldn't be using the `Digest` to do secret compare, this is why Digest is normally derived and have a non-constant `Eq`. Why are you doing this in the first place ? shouldn't you be using HMAC or some KDF maybe ?

dahlia:
I just become aware that `HMAC` also has an instance of `Eq`. Should I close this?

vincenthz:
yes, I think so. if you're doing HMAC then yes you already have a constant equality for the purpose you outlined

dahlia:
Thanks for your advice! 🙏

PR #128, opened 2017-01-09
Rename Crypto.PubKey.Ed448

ocheron:
The curve is called Ed448-Goldilocks but Ed448 also refers to EdDSA variants. The current module names in cryptonite are confusing because `Crypto.PubKey.Ed25519` is EdDSA and `Crypto.PubKey.Ed448` is D-H.

This PR renames `Crypto.PubKey.Ed448` to `Crypto.PubKey.Curve448`, with a reexport to the previous name so dependent packages have time to adjust.

If prefered, using names X25519 and X448 is also possible (but impacts two modules instead of one).

CC @centromere @kazu-yamamoto


kazu-yamamoto:
+1

vincenthz:
I think the rationale make sense to me.

It's definitely a good idea to keep old module around, byt maybe instead of adding a `DEPRECATED` right now, just add a notice that it will be DEPRECATED ~soon or something like that ? we could do couple of release without adding an explicit warning.

definitely would be nice to see if @centromere is happy with this change.

centromere:
This is a good idea. Thanks!

centromere:
@vincenthz Can this be merged now?

ocheron:
I think I'll have time to remove DEPRECATED later tonight.

ocheron:
Done: I've pushed a new version where the pragma is removed and replaced by a note in the module haddock comment.

frasertweedale:
Just a note that for consistency, module `Ed448` ought to provide EdDSA for the edwards448 curve.

Alternative, perhaps a `Crypto.Public.EdDSA` module could provide both (perhaps with a type class to present a single signing/verification interface for both edwards25519 and edwards448, and possibly others in future).

PR #132, opened 2017-02-11
Add Fast PBKDF2 for SHA1, SHA256 and SHA512

NicolasDP:
This follow up with haskell-crypto/cryptonite-openssl#4

This current implmentation is still a bit slower than @adinapoli 's [fastpbkdf2-hs](https://github.com/adinapoli/fastpbkdf2-hs) but we are still way faster than the generic implementation to make it worth adding these functions.

The following results have been generated using the benchmarks included in this PR (reporting only the 10000 iterations ones):

| Algorithm | cryptonite | fast-cryptonite | fastpbkdf2-hs |
| --- | --- | --- | --- |
| SHA1 | 30.03 ms | 6.690 ms | 4.685 ms |
| SHA256 | 21.07 ms | 8.257 ms | 5.027 ms |
| SHA512 | 26.21 ms | 10.32 ms | 6.925 ms |

### DIFF
--- a/Crypto/KDF/PBKDF2.hs
+++ b/Crypto/KDF/PBKDF2.hs
@@ -8,17 +8,22 @@
 -- Password Based Key Derivation Function 2
 --
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE ForeignFunctionInterface #-}
+
 module Crypto.KDF.PBKDF2
     ( PRF
     , prfHMAC
     , Parameters(..)
     , generate
+    , fastPBKDF2_SHA1
+    , fastPBKDF2_SHA256
     ) where
 
 import           Data.Word
 import           Data.Bits
 import           Foreign.Marshal.Alloc
-import           Foreign.Ptr (plusPtr)
+import           Foreign.Ptr (plusPtr, Ptr)
+import           Foreign.C.Types (CUInt(..), CInt(..), CSize(..))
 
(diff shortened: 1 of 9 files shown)


adinapoli:
That's fantastic news guys, well done Nicolas!

I think this is a great leap forward in terms of performance, although it requires a dive into C code. But, as Vincent said, I think the pure Haskell implementation reached its limits 😜

ocheron:
FYI compilation on the latest OpenBSD fails because of missing symbols `__builtin_bswap32` and `__builtin_bswap64` from `cbits/cryptonite_pbkdf2.c`.

Applying the solution found in https://github.com/hvr/cryptohash-sha256/commit/50e82a1acac9615573f38dca361deaf97f03707f resolves it.

Also it would be good to have documentation for the new functions, explaining what `generate` call they are equivalent to. @NicolasDP Would you like to make a new PR for it?

vincenthz:
I've removed all the write_32 / write_64 in favor of the already existing bitfn.h. it should work now

ocheron:
I confirm it works.

vincenthz:
thanks @ocheron. it would be nice to have some BSDs as part of travis to prevent this kind of things in the future (just like having some arms, mips, powerpc too :))

ocheron:
Yes it would be nice but AFAII Travis CI is linux/osx only.

PR #131, opened 2017-02-09
SHAKE128/SHAKE256 support

ocheron:
This provides support for SHAKE extendable output functions:

- to reuse `Crypto.Hash` API, but at the same time make the output length
  arbitrary, the singleton types SHAKE128/SHAKE256 are parameterized by a
  type-level `Nat` representing the digest bit size

- the C implementation extends current SHA-3 code and features a generic
  "output" function that could be called multiple times to consume output
  incrementally if needed

Feature requested in #76.

Example:
```
> :set -XDataKinds -XOverloadedStrings
> import Data.ByteString
> import Crypto.Hash
> hash ("The quick brown fox jumps over the lazy dog" :: ByteString) :: Digest (SHAKE128 256)
f4202e3c5852f9182a0430fd8144f0a74b95e7417ecae17db0f8cfeed0e3e66e
> hash ("" :: ByteString) :: Digest (SHAKE256 4096)
46b9dd2b0ba88d13233b3feb743eeb243fcd52ea62b81b82b50c27646ed5762fd75dc4ddd8c0f200
cb05019d67b592f6fc821c49479ab48640292eacb3b7c4be141e96616fb13957692cc7edd0b45ae3
dc07223c8e92937bef84bc0eab862853349ec75546f58fb7c2775c38462c5010d846c185c15111e5
95522a6bcd16cf86f3d122109e3b1fdd943b6aec468a2d621a7c06c6a957c62b54dafc3be87567d6
77231395f6147293b68ceab7a9e0c58d864e8efde4e1b9a46cbe854713672f5caaae314ed9083dab
4b099f8e300f01b8650f1f4b1d8fcf3f3cb53fb8e9eb2ea203bdc970f50ae55428a91f7f53ac266b
28419c3778a15fd248d339ede785fb7f5a1aaa96d313eacc890936c173cdcd0fab882c45755feb3a
ed96d477ff96390bf9a66d1368b208e21f7c10d04a3dbd4e360633e5db4b602601c14cea737db3dc
f722632cc77851cbdde2aaf0a33a07b373445df490cc8fc1e4160ff118378f11f0477de055a81a9e
da57a4a2cfb0c83929d310912f729ec6cfa36c6ac6a75837143045d791cc85eff5b21932f23861bc
f23a52b5da67eaf7baae0f5fb1369db78f3ac45f8c4ac5671d85735cdddb09d2b1e34a1fc066ff4a
[...]

PR #134, opened 2017-02-14
SHAKE128 / SHAKE256 support

vincenthz:
supersede #131


Crypto/Hash/SHAKE.hs
```diff
@@ -0,0 +1,91 @@
...
+    hashInternalFinalize      = shakeFinalizeOutput (Proxy :: Proxy bitLen)
+
+shakeFinalizeOutput :: (IsDivisibleBy8 bitLen, KnownNat bitLen)
+                    => proxy bitLen
+                    -> Ptr (Context a)
+                    -> Ptr (Digest a)
+                    -> IO ()
+shakeFinalizeOutput d ctx dig = do
+    c_sha3_finalize_shake ctx
+    c_sha3_output ctx dig (fromInteger (natVal d `div` 8))
```
ocheron, line 76: Any reason for not using function `byteLen` ?

Crypto/Hash/SHA3.hs
```diff
@@ -6,7 +6,7 @@
 -- Portability : unknown
 --
 -- module containing the binding functions to work with the
--- SHA3 cryptographic hash.
+-- SHA3 cryptographic hash and extendable output functions.
```
ocheron: This file should be reverted entirely.

Crypto/Hash/SHAKE.hs
```diff
@@ -0,0 +1,91 @@
...
+
+-- | SHAKE128 (128 bits) extendable output function.  Supports an arbitrary
+-- digest size (multiple of 8 bits), to be specified as a type parameter
+-- of kind 'Nat'.
+--
+-- Note: outputs from @'SHAKE128' n@ and @'SHAKE128' m@ for the same input are
+-- correlated (one being a prefix of the other).  Results are unrelated to
+-- 'SHAKE256' results.
+data SHAKE128 (bitlen :: Nat) = SHAKE128
+    deriving (Show, Typeable)
```
ocheron, line 41: Deriving `Data` is missing for SHAKE128 and SHAKE256

ocheron:
OK for using a separate module and conditional compilation. The latter will be a necessary evil when it comes to dependencies from other packages. I guess this situation will be only "temporary".

About generalization of SHA3 output length to any Nat: only specific sizes have been standardized, so I think it would be more confusing than useful. For Stein / Keccak: why not.

About what happens when the bit length is not divisible by 8: I don't view it so problematic. FIPS 202 defines an encoding for any bit size, even when not aligned on full bytes. In the examples there is [one with partial bytes](http://csrc.nist.gov/groups/ST/toolkit/documents/Examples/ShakeTruncation.pdf). I just did not take time to implement it. It's an optional feature and I think it will much less popular.

So I have no objection if you want to keep the `IsDivisibleBy8` type constraint, it's nicely done. But the justification for all the language extensions involved is rather weak at this point. I can implement the truncation if you prefer instead.

Btw I'm also a bit worried about the documentation in `GHC.TypeLits`, explaining the module should not be used directly but through a library instead. This library does not seem to exist yet.

In crypto code we have lots of bytearrays with length known at compile time, so there is probably room for a library providing necessary utilities. Something like constraints with more expressivity than ByteArray / ByteArrayAccess.

vincenthz:
I haven't looked at FIPS 202 bit size encoding, but if the scheme is defined at the bit level it make sense to put on the constraint on the bit size that are actually supported right now. if/once we get the bitsize support, then we can remove the constraint.

Related to the extensions:

* I don't see it too much of an issue going forward, the needed extensions are actually not too bad, mostly TypeFamilies and DataKind, some of them are somewhat optional like TypeError and ConstraintKind.
* We could support IsDivisibleBy8 as a dummy constraint for compiler that doesn't have the needed extension. it wouldn't actually provide the real constraint, but at least you'll be able to write the constraint for future compat. Maybe a moot issue provided that older compiler don't support `Nat` anyway

I've started thinking about adding sized vector in `foundation` already, so maybe this would be useful there. I'm actually thinking of adding the `IsDiv8` constraint in foundation, and some other related things to what I played with coming up with IsDivisibleBy8.

ocheron:
OK, thanks. That will be an example of what can be done.

ocheron:
For information I pushed a modification directly to master so that the test suite runs with 7.6.3.
A bit too late for the new release unfortunately.

vincenthz:
@ocheron if you want we can do a quick .1 release to fix this issue ?

ocheron:
It's not like I need it, but if you think a new relase will be useful for someone, then yes of course.
What are the steps?

PR #138, opened 2017-02-20
Blake2 & Argon2

vincenthz:
* Update blake2 to latest version
* Hide X448_Base_point from external symbols
* Add a modified version of argon2:
  * hide all internal symbols
  * prefix all external symbol with `cryptonite_`


vincenthz:
There's now a working Argon2 hash function. I change things quite a lot to make it more cryptonite compliant:

* It doesn't rely on the C error values, instead trying to verifying all the parameter ahead of time (not complete) and using the CryptoFailable facility to report the error.
* it use the `ByteArray` / `ByteArrayAccess` classes so that we can fill safer kind of buffer that are auto-erasable and have time constant equality (e.g `ScrubbedBytes`).
* It doesn't have all the encoding string business, remove lots of C code, and I'ld rather have some format printing & base64 in haskell.
* All the verification code is removed too, this is just a facility that take back the encoded string, fill the options structure, run the hash and do a safe verification. I'ld rather do that on haskell too.

PR #140, opened 2017-03-13
Blake2: add modern interface using Nat

NicolasDP:
Just continuing the improvements started with `Cryptonite.Hash.SHAKE`.


ocheron:
BLAKE2 is defined with a maximum digest size.
Is it possible to add this contraint using the operators from `GHC.TypeLits` ?

NicolasDP:
That's a good idea. I will see if I can come with something in `Crypto.Internal.Nat`.

NicolasDP:
Now, with `base >= 4.9.0`, we can have the nice following output:

```
/Users/nicolas/work/primetype/haskell-crypto/cryptonite/tests/Hash.hs:202:23: error:
    • bitlen 1024 is greater than 512
      You have tried to use an invalid Digest size. Please, refer to the documentation.
    • In the expression: HashAlg (Blake2b :: Blake2b 1024)
      In the expression:
        ("Blake2b 512", HashAlg (Blake2b :: Blake2b 1024), 
      ...
```

Also added the minimal constraint of 1 byte.

vincenthz:
Nice, thanks !

PR #141, opened 2017-03-17
Update Blake2b.hs

alexanderkjeldaas:
Document which is the standard Blake2

### DIFF
--- a/Crypto/Hash/Blake2b.hs
+++ b/Crypto/Hash/Blake2b.hs
@@ -21,7 +21,7 @@ import           Data.Typeable
 import           Data.Word (Word8, Word32)
 
 
--- | Blake2b (160 bits) cryptographic hash algorithm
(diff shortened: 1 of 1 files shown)


tekul:
Could these changes not be made in a single commit, rather than three separate PRs? Also, as someone who isn't familiar with the code, how does this change "document which is the standard blake2" ? A clearer explanation would be nice.
 
Also the main commit message should preferably make it clear that this is only a doc change - "Update file x" isn't very helpful when you are reviewing the commit log to find relevant changes for retrospective patches.

vincenthz:
I agree with the multiples PR, that doesn't warrant it. also those files are autogenerated from `gen`, so I'ld rather not change anything inside in it, and there really should be a strong incentive to modify them.

Also I like the precision in the documentation, if it could make to work with the gen infrastructure that would perfect

alexanderkjeldaas:
I'm just editing on github, and the three PRs is what github is giving me.

PR #145, opened 2017-03-19
Tutorial Improvement

tdietert:
Please suggest other examples to implement if desired. The type signatures of the functions are very general and am unsure of how to best present the examples, but this compiles out of the box and the functions defined have the nice feature of working for most (?) symmetric block ciphers. 

Note: I realize I didn't follow the contribution guidelines (error throwing and indentation), but assumed it doesn't matter as much because this is a tutorial module and not source code for the library. I will be happy to change it if preferred.

### DIFF
--- a/Crypto/Tutorial.hs
+++ b/Crypto/Tutorial.hs
@@ -1,9 +1,10 @@
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE GADTs #-}
 
 {- How to use @cryptonite@ with symmetric block ciphers
 
+> {-# LANGUAGE OverloadedStrings #-}
+> {-# LANGUAGE ScopedTypeVariables #-}
+> {-# LANGUAGE GADTs #-}
+> 
 > import           Crypto.Cipher.AES (AES256)
 > import           Crypto.Cipher.Types (BlockCipher(..), Cipher(..), nullIV, KeySizeSpecifier(..))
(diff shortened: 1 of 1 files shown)


tekul:
Thanks for the contribution. This is a big improvement over the previous version which didn't have a fully random key.

One suggestion I would make is about the use of the function `genPrivateKey`. "Private key" might be confused with public key crypto, so perhaps "secret key" would be a better term for symmetric encryption. Also the function is quite scary looking when all it is really doing is generating a string of bytes of a particular length. Since the `exampleAES256` function is already in IO, I would just generate the key where it is used, rather than supplying a generic function. That way people have the simplest example possible.

Another flaw in the current version is that it uses `nullIV`. In most cases people will also want to generate an IV when using AES. I think that should also be part of the example.

tdietert:
Sure! Thanks for your thoughts, I think all your suggestions are warranted. :) I'll make those changes tonight.

tdietert:
@tekul I have updated the tutorial based on your suggestions, though I don't fully agree that the generic functions are scary looking-- they provide functions that user can use for any crypto type with an instance of `BlockCipher`, by simply copying and pasting. The code in `exampleAES256` should be enough for new users to understand how to use these functions and what they are doing, but I understand if you disagree and would accept further suggestions. Regardless, I feel like this is a great improvement from the original "tutorial"!

vincenthz:
Thanks, this is looking good. I'm on the fence with the generic functions (in term of simplicity), but this is probably also useful to demonstrate how to use things more generically too.

ocheron:
Haddock syntax now fixed in commit 554f0fc.

PR #149, opened 2017-04-04
Hash benchmarks

ocheron:
This adds benchmarks for the same list of algorithms than in the test suite.

I've selected a very small input size to test the overhead of the hash API itself and a larger size which should be more representative of the algorithms.

### DIFF
--- a/benchs/Bench.hs
+++ b/benchs/Bench.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE PackageImports #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ExistentialQuantification #-}
 module Main where
 
 import Criterion.Main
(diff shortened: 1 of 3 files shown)


benchs/Bench.hs
```diff
@@ -14,15 +15,69 @@ import           Crypto.Hash
...
 benchHash =
-    [ 
+    [ bgroup "1KB" $ map (doHashBench oneKB) hashAlgs
+    , bgroup "1MB" $ map (doHashBench oneMB) hashAlgs
     ]
+  where
+    doHashBench b (name, HashAlg alg) = bench name $ nf (hashWith alg) b
+
+    oneKB :: Bytes
+    oneKB = unsafePerformIO (getRandomBytes 1024)
```
vincenthz, line 37: not sure there's any point getting random; an hash algorithms behave exactly the same whether it's 0s or 1s or random data. (i.e. you'ld do as well by using `"replicate 1Mb '0'"`)

vincenthz (approved):
that's quite useful to be able to bench one of the critical part of cryptonite :+1: I'm happy to see this revived.

vincenthz:
If you want to merge without addressing the 0 vs random go ahead

benchs/Bench.hs
ocheron: Yeah, that's me being extra cautious just in case there is a bias with some simple sequence. And also my lack of imagination about what to choose instead.

So I'll merge with random but anyone more convinced can change it.

PR #157, opened 2017-05-02
ed25519: Adding generateSecretKey and a unit test

ixmatus:
This change adds a function for generating an Ed25519 secret key and exports the size constants for the `PublicKey`, `SecretKey`, and `Signature` for use with `getRandomBytes` and other functions.

The motivation for this change came from a CLI tool I wrote to generate an ed25519 key pair and sign and verify messages with ed25519 key pairs. I noticed Curve25519 had a generateSecretKey function but the ed25519 module did not so I wrote the code to generate it myself using `getRandomBytes`. I also goofed the first implementation by supplying a byte size to `getRandomBytes` that was greater than the number expected by the `Ed25519.secretKey` constructor so I attempted to use the `secretKeySize` function but it was not exported.

@vincenthz I noticed the `Curve25519.generateSecretKey` function modifies some of the generated random bytes; I don't understand the reason for it and instead of copying and pasting that code I figured I'd assume it was specific to `Curve25519` and leave it out and if that assumption is wrong, be schooled by you ;)

### DIFF
--- a/Crypto/PubKey/Ed25519.hs
+++ b/Crypto/PubKey/Ed25519.hs
@@ -7,12 +7,17 @@
 --
 -- Ed25519 support
 --
+{-# LANGUAGE BangPatterns               #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
 module Crypto.PubKey.Ed25519
     ( SecretKey
     , PublicKey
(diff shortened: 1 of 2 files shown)


vincenthz:
Thanks !

Related to the key modification in curve25519/ed25519, you're right to assume that it is not the case in ed25519. ed25519 hashes its "secret key" first before masking the result, whereas curve25519 is already in a usable secret key form which need its bits trimmed to prevent catastrophic timing failure.

PR #158, opened 2017-05-10
HashBlockSize, HashDigestSize, HashInternalContextSize as type functions

int-index:
I kept changes to the minimum and tried to preserve backwards-compatibility.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 gen/Gen
 gen/Crypto/*
 dist
+dist-newstyle
 QA
 benchs/Bench
 benchs/Hash
--- a/Crypto/Hash/Blake2b.hs
+++ b/Crypto/Hash/Blake2b.hs
@@ -10,6 +10,8 @@
 --
 {-# LANGUAGE ForeignFunctionInterface #-}
 {-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE TypeFamilies #-}
 module Crypto.Hash.Blake2b
     (  Blake2b_160 (..), Blake2b_224 (..), Blake2b_256 (..), Blake2b_384 (..), Blake2b_512 (..)
     ) where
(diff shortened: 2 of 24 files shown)


vincenthz:
Kudos on keeping it compatible. However the thing that bothers me is that we would be stuck with the `HashAlgorithm'` class which express something quite fundamental but at the same time optional

So for example `Crypto.Hash.Skein` and `Crypto.Hash.Blake` were not updated in this PR (which is fine), but shows somewhat the problem I'm trying to highlight, you can define HashAlgorithm without HashAlgorithm'

Adding @ocheron @NicolasDP @tekul @centromere  to have further opinions.

Couple of options:
* Get rid of hashBlockSize/hashDigestSize/hashInternalContextSize in the class itself, and have a `Nat` friendly version + a outside-of-class functions to keep compat. Potentially can break compatibility,
* Crypto.Hash2 to define a brand new Crypto.Hash class with the right methods and none of the crappy `undefined :: a` style one.
* Go with this
* Other ?

centromere:
What are the benefits of making this change?

tekul:
I don't think I understand what this is doing, so I probably can' t offer a useful opinion unless someone can explain it to me. Maybe you can explain some of the recent type and compiler error stuff to me at Zurihac, Vincent :-).

vincenthz:
sorry @tekul and @centromere, for the lack of context.

One of the issue is that the interface of `HashAlgorithm` is quite old and is defined (just like `Storable`):

```
hashBlockSize :: hashAlg -> Int
hashBlockSize _ = 12
```

that present the annoying issue that you might need to use it with `undefined` to be able to generically use it:

```
hashBlockSize (undefined :: SHA1)
```

so this solve 2 problems:

* remove the need for undefined when calling one of those 3 functions
* define the length of a hash at the type level (thus as a `Nat`)

happy to explain everything at the zurihac @tekul :)

centromere:
Couldn't you avoid `undefined` by using `Data.Proxy`?

int-index:
> Couldn't you avoid undefined by using Data.Proxy?

Yes, you could, and that'd be an improvement. However, having those values on type level gives you this and extra. My use case was using `HashDigestSize` as a type parameter to the `StaticSize` type constructor in the `store` serialization library.

NicolasDP:
I think it would be better with the length of the hash at the type level.

One of the upside is it allows us to have less custom `type` to define. Right now, if we want to add a new digest size for every new digest one may need to use. This is a bit of a shame really.
In #140 I have been able to define all the different Blake2 digest sizes. One may want to generate a `Blakeb` digest size of 1 byte: it is possible now (a bit useless... though) and no need to make a PR to add `Blake2b_1`.

The downside is that we will break compatibility with the older compilers (we will need at least `ghc >= 7.8`).

ocheron:
The solution with new modules is attractive because it can give new API and implementation using it right away. However I'm afraid migrating to this means new modules HMAC2, KDF2, etc. so it goes probably too far.

Also I'm not sure if the Proxy part and Nat associated types must be related. For example it's possible to add hashDigestSize' to HashAlgorithm with a default implementation based on hashDigestSize.

vincenthz:
hm this is a good point about HMAC2, KDF2. clearly not practical.

Now I'm thinking the first option might be better. While potentially, someone could have defined its own HashAlgorithm, I wouldn't be surprised if no-one actually did. Updating `HashAlgorithm` directly should bring really small breakage.

vincenthz:
thanks @int-index for starting this; I ended up merging something slightly different, not using a different class but instead improving HashAlgorithm itself in the same way as this PR. in couple of releases we can deprecated the functions that get the various sizes, and in further release remove the functions altogether.

PR #159, opened 2017-05-16
Allow sign/verify digest directly

wangbj:
currently sign/verify works on message directly, it would be nice if PSS could sign/verify digest directly. This is useful for:

  1) for some signing server it only has a digest (without message)
  2) message could be very large, for cases when client need request a singing server to sign, it may make more sense for the client to compute digest, then ask server to (PSS) sign the digest
  3) openSSL pkeyutl (PSS) sign operation signs with digest only, not the message, it would be nice to work with openSSL more easily 

*openSSL command line:
```shell
openssl pkeyutl -pkeyopt rsa_padding_mode:pss -pkeyopt rsa_pss_saltlen:-1  -pkeyopt digest:sha256 -sign -inkey "pri.key" -in hmac.bin > sig.bin
openssl pkeyutl -pkeyopt rsa_padding_mode:pss -pkeyopt rsa_pss_saltlen:-1  -pkeyopt digest:sha256 -verify -inkey "pri.key" -in hmac.bin -sigfile sig.bin
```

I also created a test case to test PSS sign/verify interop with OpenSSL:

https://github.com/wangbj/sign-pss-test/blob/master/test/Spec.hs

### DIFF
--- a/Crypto/PubKey/RSA/PSS.hs
+++ b/Crypto/PubKey/RSA/PSS.hs
@@ -66,13 +66,14 @@ signDigestWithSalt :: HashAlgorithm hash
                    -> Maybe Blinder -- ^ optional blinder to use
                    -> PSSParams hash ByteString ByteString -- ^ PSS Parameters to use
                    -> PrivateKey    -- ^ RSA Private Key
-                   -> ByteString    -- ^ Message digest
+                   -> Digest hash   -- ^ Message digest
                    -> Either Error ByteString
-signDigestWithSalt salt blinder params pk mHash
+signDigestWithSalt salt blinder params pk digest
     | k < hashLen + saltLen + 2 = Left InvalidParameters
     | hashLen /= B.length mHash = Left InvalidParameters
     | otherwise                 = Right $ dp blinder pk em
     where k        = private_size pk
+          mHash    = B.convert digest
           dbLen    = k - hashLen - 1
           saltLen  = B.length salt
           hashLen  = hashDigestSize (pssHash params)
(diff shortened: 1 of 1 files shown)


ocheron:
Did you consider changing the `ByteString` input to `Digest hash` ?
The signature would be more expressive and we don't need to test the input length anymore (as the constraint is enforced by the data type).

wangbj:
Good idea, have made the change. Do you think we should allow ``signDigest`` on other RSA module (such as pkcs15)?

ocheron:
Good but when signing we don't need to test `hashLen /= B.length mHash` anymore.
Can you remove this condition too?

> Do you think we should allow signDigest on other RSA module?

Only if someone actually needs it.

cl04:
Well, I have an application could use it, but guess it's hard to justify with limited usage :) I think the reasons listed before still stands.

cl04:
sorry by using another account

vincenthz:
Looking good, specially after @ocheron 's suggestion !

If there's the same limitation in pkcs15 RSA, then yes we could eventually lift this limitation with the same approach

wangbj:
It seems PKCS15 doesn't really need any change, it allows passing ``Nothing::Maybe hashAlg`` to skip the hashing step. PSS doesn't support this, though

ocheron:
I don't think it's exactly the same as PKCS15 includes a prefix inside this value.
I'll be happy to merge this version or wait if you want to generalize.

wangbj:
Do you think it necessary to introduce a sign/verify Digest function to all asymmetric algorithms? I'd like to do so, but it will take more time.

ocheron:
No I don't think it's necessary.
With RSA PSS as example, it's always possible to add more signature schemes later.
So I merge this part for now.

PR #161, opened 2017-05-17
[ChaChaPoly1305] Add Eq and Ord instances for Nonce

centromere:
The ability to compare Nonces is important because it allows you to detect when you reach the maximum nonce, thus preventing overflow resulting in catastrophic key re-use.

### DIFF
--- a/Crypto/Cipher/ChaChaPoly1305.hs
+++ b/Crypto/Cipher/ChaChaPoly1305.hs
@@ -84,14 +84,12 @@ instance ByteArrayAccess Nonce where
 instance Eq Nonce where
   (Nonce8  a) == (Nonce8  b) = a == b
   (Nonce12 a) == (Nonce12 b) = a == b
-  (Nonce8  a) == (Nonce12 b) = a == b
-  (Nonce12 a) == (Nonce8  b) = a == b
+  _           == _           = error "unable to compare 8-byte and 12-byte nonces"
 
 instance Ord Nonce where
   compare (Nonce8  a) (Nonce8  b) = compare a b
   compare (Nonce12 a) (Nonce12 b) = compare a b
-  compare (Nonce8  a) (Nonce12 b) = compare a b
-  compare (Nonce12 a) (Nonce8  b) = compare a b
+  compare _           _           = error "unable to compare 8-byte and 12-byte nonces"
 
 -- Based on the following pseudo code:
 --
--- a/Crypto/Cipher/ChaChaPoly1305.hs
+++ b/Crypto/Cipher/ChaChaPoly1305.hs
@@ -81,6 +81,18 @@ instance ByteArrayAccess Nonce where
   withByteArray (Nonce8  n) = B.withByteArray n
   withByteArray (Nonce12 n) = B.withByteArray n
 
+instance Eq Nonce where
+  (Nonce8  a) == (Nonce8  b) = a == b
+  (Nonce12 a) == (Nonce12 b) = a == b
+  (Nonce8  a) == (Nonce12 b) = a == b
+  (Nonce12 a) == (Nonce8  b) = a == b
+
+instance Ord Nonce where
+  compare (Nonce8  a) (Nonce8  b) = compare a b
+  compare (Nonce12 a) (Nonce12 b) = compare a b
+  compare (Nonce8  a) (Nonce12 b) = compare a b
+  compare (Nonce12 a) (Nonce8  b) = compare a b
+
 -- Based on the following pseudo code:
 --
 -- chacha20_aead_encrypt(aad, key, iv, constant, plaintext):


centromere:
I am going to be needing this soon for a library of mine. @vincenthz are you able to take a look?

Crypto/Cipher/ChaChaPoly1305.hs
```diff
@@ -81,6 +81,18 @@ instance ByteArrayAccess Nonce where
   withByteArray (Nonce8  n) = B.withByteArray n
   withByteArray (Nonce12 n) = B.withByteArray n
 
+instance Eq Nonce where
+  (Nonce8  a) == (Nonce8  b) = a == b
+  (Nonce12 a) == (Nonce12 b) = a == b
+  (Nonce8  a) == (Nonce12 b) = a == b
```
vincenthz: instead of checking mismatching constructor, why not doing a `_ == _ = False`  directly ?

Crypto/Cipher/ChaChaPoly1305.hs
```diff
@@ -81,6 +81,18 @@ instance ByteArrayAccess Nonce where
...
+  (Nonce12 a) == (Nonce8  b) = a == b
+
+instance Ord Nonce where
+  compare (Nonce8  a) (Nonce8  b) = compare a b
+  compare (Nonce12 a) (Nonce12 b) = compare a b
+  compare (Nonce8  a) (Nonce12 b) = compare a b
```
vincenthz: in a case of mismatch of constructor, I'm not sure what we're really comparing and whether is make sense to compare a Nonce12 and a Nonce8 (and what the result mean)

Crypto/Cipher/ChaChaPoly1305.hs
centromere: That seems reasonable. I will update.

Crypto/Cipher/ChaChaPoly1305.hs
vincenthz: further thoughts, would it make sense to just consider `Nonce12` to be by construction bigger than `Nonce8` maybe ? (i.e. irrelevant of the data inside)

centromere: Both Nonce8 and Nonce12 are 12 bytes under the hood, so the comparison does "work" in some sense. I admit that I'm not happy with it, but I don't have a better solution.

centromere: > would it make sense to just consider Nonce12 to be by construction bigger than Nonce8 maybe

I don't think so. If the 4 byte "constant" is all zeros, a Nonce12 can be smaller than a Nonce8.

vincenthz: when they are considered as integer yes, but here I'm saying to basically consider them as 2 independent spaces. It's probably a good idea to be consistent with == where
`Nonce8 0000 0000` would be not equal to `Nonce12 0000 0000 0000`. if you compare the byte in `Ord`, then you run into the problem of having something that is not `==` but `compare == EQ` hold

centromere: Ok sure. I am not sure how to handle that though.

vincenthz:
sorry just read the description, so now I understand why you want this; but I'm not sure that make sense.

overflow should happens potentially after 2^64 chacha encryption (of each 64 bytes), so that's a million petabytes before nonce overflow. I don't remember off-hands, but I think you're suppose to rekey your chacha cipher much much earlier than this.

edit: I'm still ok with adding `Eq` and `Ord` instance to Nonce. Also if you really want to compare nonce, then that's fine too, although I wonder if potentially you can create a timing skew everytime you do that and potentially let someone guess your nonce (if you're in the rare case where you keep the nonce private)

Crypto/Cipher/ChaChaPoly1305.hs
vincenthz: if you consider one greater than the other just by the constructor then effectively there's no way for a Nonce8 to be greater than a Nonce12 and then you're effectively in two different compare space.

vincenthz: Also means that there's no way that a Nonce8 compared to a Nonce12 returns a Eq this way.

centromere:
Alternatively, I could (in my application) store a separate Integer containing the number of times the Nonce has been incremented, but that feels dirtier than doing it this way.

centromere:
@vincenthz Do you need more information before making a decision one way or the other on this?

vincenthz:
I'm not sure how you want to proceed here. Do you want to update the Eq/Ord to make Nonce8 and Nonce12 completely different or not (in a sense of not comparing them with their content) ? I think you mentioned that you would update `Eq`, but I haven't seen anything new

centromere:
I am not sure how to handle an equality comparison between `Nonce8` and `Nonce12`. Since I can't seem to decide, I suppose it might make sense to just close this PR.

centromere:
Would it be acceptable to simply error when the incompatible nonces are compared, as I did in my latest commit?

centromere:
It just occurred to me that Nonce derives ByteArrayAccess, so I can do the comparison on my own without these instances.

NicolasDP:
good point, it's a `ByteArrayAccess`... The underlying type is a `Bytes` from `memory`. @vincenthz , since you said it could be a security issue to allow comparison of the `Nonce`, would it make sense to make it a [ScrubbedBytes](http://hackage.haskell.org/package/memory/docs/Data-ByteArray.html#t:ScrubbedBytes) to at least have the constant time equality ?

Issue #160, opened 2017-05-17
Unexpected ivAdd overflow behavior


centromere:
When testing the [`ivAdd`](https://www.stackage.org/haddock/lts-8.13/cryptonite-0.21/Crypto-Cipher-Types.html#v:ivAdd) function I noticed that when intentionally given a value of all 0xFF bytes, it rolls over to `0x00...0x01`. My expectation was that it would roll over to `0x00...0x00`. Is this intended behavior?

> λ> import Data.Maybe
> λ> import Data.ByteArray
> λ> import Data.ByteString
> λ> import Crypto.Cipher.Types
> λ> import Crypto.Cipher.AES
> λ> let n = fromJust . makeIV $ ("\255\255\255\255\255\255\255\255\255\255\255\255\255\255\255\255" :: ByteString) :: IV AES256
> λ> Data.ByteString.unpack (convert n :: ByteString)
> [255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255]
> λ> Data.ByteString.unpack $ (convert $ ivAdd n 1 :: ByteString)
> [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1]

ocheron:
In the implementation the carry is added back to the least significant byte.

This could just be a bug because ivAdd is used by ctrCombine, in turn demonstrated in Crypto.Tutorial.

If I set the IV to 0xffffffffffffffffffffffffffffffff or very close, this does not interoperate
with `openssl enc -aes-256-ctr`. Whereas when the IV is random, one implementation
can decrypt what the other encrypted.

vincenthz:
yeah that looks like a bug indeed.

ocheron:
Fixed with #164.

PR #164, opened 2017-05-30
Fix ivAdd overflow behaviour

ocheron:
Following the discussion in #160, I confirm the code we have today can cause issues when the IV is all 0xFF bytes or close. This becomes visible with AES CTR as soon as I remove the C implementation of ctrCombine and use the default ctrCombineGeneric, based on ivAdd.

### DIFF
--- a/Crypto/Cipher/Types/Block.hs
+++ b/Crypto/Cipher/Types/Block.hs
@@ -36,7 +36,6 @@ module Crypto.Cipher.Types.Block
     --, cfb8Decrypt
     ) where
 
-import           Control.Monad (unless)
 import           Data.Word
 import           Data.Monoid
 import           Crypto.Error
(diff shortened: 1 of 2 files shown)


Crypto/Cipher/Types/Block.hs
```diff
@@ -167,24 +168,16 @@ nullIV = toIV undefined
...
-                  case r of
-                      0 -> return ()
-                      _ -> until0 r
-            until0 i
+        copy bs = B.copyAndFreeze bs $ loop i (B.length bs - 1)
 
-        loop :: Int -> Int -> Ptr Word8 -> IO Int
-        loop 0   _   _ = return 0
+        loop :: Int -> Int -> Ptr Word8 -> IO ()
+        loop 0   _   _ = return ()
```
vincenthz: we could get rid of this line to make it more constant-time to `ivAdd`.

vincenthz (approved):
looking good from my end. happy to merge as is, but I made a note about the iv non-constant algorithm.

ocheron:
Indeed it's an easy gain.
I also pattern match on `ofs` to get rid of `unless` and make `ofs >= 0` more explicit.

PR #166, opened 2017-06-01
add aes ccm (RFC3610) support?

wangbj:
do you plan to support aesccm defined in RFC3610? aesccm is less flexible and is slower than gcm. This pull request implements aesccm, with below limitations:

* AEAD_CCM constructor has is ``AEAD_CCM input_size M_size L_size``
It seems aesccm first block (`b0`) depends on input message size, the cbcmac depends on b0, thus I didn't find a better way to do this, unless we save the entire aad, because aad is inserted before encryption.

We can get rid of M_size and L_size, such as always set M_size to 16 (cbcmac length), and L_size based on input_size. I kept this for now to pass all the tests in rfc3610 examples (M_size is 10 or 8).

* since AEAD_CCM constructor has message size, may want to throw crypto error when ccmEncrypt pass input buffer with a different size, instead of assertion failure.

* max L_size is 4, so maximum message size is 2^32, even though aesccm standard supports message size up to 2^64 (2^(8*L)), it seems impractical to do so, and some aesccm implementation supports L=2 only (thus maximum message size is 2^16). This also implies `aad` maximum size is (2^32)

* the `nonce` size is `15-L_size`, this implementation truncates `nonce` silently if it is longer than `15-L_size`, however, it maybe better to throw error instead.


wangbj:
I realized AEAD_CCM constructor isn't quite idiomatic, so does CCM_Mx/CCM_Lx, however it can be can be improved by always trying to use the longest nonce whenever possible, and always assume M=16. What do you think?

ocheron:
Yes CCM block B_0 does not fit well with the Haskell incremental API.
Message and tag lengths are known only at the end, when `aeadFinalize` is called.
And it's not even possible to return a CryptoFailed result if the actual values do not match what you set in the constructor.

So with the current API I'd say we need to save everything and defer authentication-tag computation to the very end. And keep L very small too.

RFC 6655 defines TLS ciphers where authentication tag is 8 or 16 bytes but I don't think `tls` can manage 8 bytes currently.

Waiting other's opinion. I don't know symmetric crypto very well.

vincenthz:
nice work; the only thing is missing is a handful of test vector to make sure we're actually implementing the real deal, please add some

wangbj:
Thanks, do you think it's better to support different M/L size (at the cost of cluttering the API), or support constant M/L (16/13, at the cost of support maximum message size of 64K, as some/most implementations)?

ocheron:
@wangbj Since you implemented the general solution with any M and L values why not keep it that way.

I've been looking how to remove the message length `n` from the AEAD_CCM constructor but indeed this means deferring AAD processing, and would prevent feeding message blocks incrementally if we need that later. In the end I don't think this makes much of a difference for usability.

Vectors from RFC 3610 are in ocheron@bb50935 and I also added `tls` ciphers to test interoperability. Everything works OK 👍.

cbits/cryptonite_aes.c
```diff
@@ -406,6 +438,139 @@ void cryptonite_aes_gcm_finish(uint8_t *tag, aes_gcm *gcm, aes_key *key)
...
+	} else if (la < (1ull << 32)) {
+		output->b[0] = 0xff;
+		output->b[1] = 0xfe;
+		output->b[2] = (la >> 24) & 0xff;
+		output->b[3] = (la >> 16) & 0xff;
+		output->b[4] = (la >>  8) & 0xff;
+		output->b[5] = la         & 0xff;
+		return 6;
+	}
+}
```
ocheron, line 478: With clang I get two warnings related to function `ccm_encode_la`:
```
cbits/cryptonite_aes.c:470:16: error:
     warning: comparison of constant 4294967296 with expression of type 'unsigned int' is always true [-Wtautological-constant-out-of-range-compare]
            } else if (la < (1ull << 32)) {
                       ~~ ^ ~~~~~~~~~~~~
    |
470 |         } else if (la < (1ull << 32)) {
    |                ^

cbits/cryptonite_aes.c:479:1: error:
     warning: control may reach end of non-void function [-Wreturn-type]
    |
479 | }
    | ^
}
^
2 warnings generated.
```

Can variable `la` be changed to `uint32_t` (like in the caller) and the test removed?

cbits/cryptonite_aes.h
```diff
@@ -55,6 +55,18 @@ typedef struct {
...
+	aes_block xi;
+	aes_block header_cbcmac;
+	aes_block b0;
+	aes_block nonce;
+	unsigned length_aad;
+	unsigned length_input;
+        int      length_M;
+        int      length_L;
+} aes_ccm;
+
```
ocheron: I think it's best not to use types with variable size like `int` or `unsigned` but fixed-sized like `uint8_t` and `uint32_t`.

cbits/cryptonite_aes.c
```diff
@@ -406,6 +438,139 @@ void cryptonite_aes_gcm_finish(uint8_t *tag, aes_gcm *gcm, aes_key *key)
...
+
+/* even though it is possible to support message size as large as 2^64, we support up to 2^32 only */
+void cryptonite_aes_ccm_init(aes_ccm *ccm, aes_key *key, uint8_t *nonce, uint32_t nonce_len, uint32_t input_size, int m, int l)
+{
+	memset(ccm, 0, sizeof(aes_ccm));
+
+	if (l < 2 || l > 4) return;
+	if (m != 4 && m != 6 && m != 8 && m != 10
+		   && m != 12 && m != 14 && m != 16) return;
```
ocheron, line 505: Testing `l` and `m` values could even be an assertion since Haskell has ADTs.

cbits/cryptonite_aes.c
```diff
@@ -406,6 +438,139 @@ void cryptonite_aes_gcm_finish(uint8_t *tag, aes_gcm *gcm, aes_key *key)
...
+
+	if (nonce_len != 15 - l) {
+		nonce_len = 15 - l;
+	}
```
ocheron: Not sure what to do with this.

We could ensure `nonce_len >= 15 - l` (ignoring extra bytes) and move the test to Haskell so `ccmInit` returns a failure when `nonce_len` is too short.

Function `cryptonite_aes_ocb_init` already has something like this:
```c
        /* we don't accept more than 15 bytes, any bytes higher will be ignored. */
        if (len > 15) {
                len = 15;
        }
```

ocheron (commented):
I spotted a few places where the C code could be made more strict, please tell me what you think.

cbits/cryptonite_aes.h
wangbj: @ocheron Many thanks for your through review I agree with your about the changes, also there're three asserts in the change I think they better be removed.

cbits/cryptonite_aes.c
wangbj: Should it be changed to if (nonce_len >= 15 - l) nonce_len = 15 -l instead? Because ccmInit returns a concrete type (instead of Maybe/Either), it isn't straight forward to handle errors unless throw exceptions. Do you have any suggestions so that errors can be handle more gratefully?

ocheron: Here's how I view things:
1. at C level priority is to ensure we don't read/write outside allocated buffers. So the `memcpy` call can be changed too, so that `nonce_len` is now used
2. in Haskell `ccmInit` and `aeadInit` can return CryptoError_IvSizeInvalid when the size is not compatible with L

ocheron:
@wangbj Any reason for not putting the length check directly in`ccmInit` so it applies also to the `BlockCipher` instance defined in `Crypto.Cipher.AES.Primitive`?

wangbj:
@ocheron ccmInit returns AESCCM, so only way ccmInit for ccmInit return CryptoError is by throwing exceptions?

ocheron:
Yes so why not changing the type to `CryptoFailable AESCCM`?
The function is not exposed, only used internally.

ocheron (approved):
That's good, I rebase and merge this with a few style improvements as well as other changes that have been applied to master.

Issue #84, opened 2016-04-27
Improve Ed448 implementation


centromere:
The current Ed448 implementation was taken from [here](https://sourceforge.net/p/ed448goldilocks/code/ci/x448/tree/), however an improved version with better performance is located [here](https://sourceforge.net/p/ed448goldilocks/code/ci/curve25519-work/tree/).

Unfortunately, due to the way libdecaf is structured, extracting only the x448 operation is not trivial. It is not neatly contained within two files.

vincenthz:
I had a bit of a look, but it is definitely non-trivial to do

ocheron:
Has anyone continued working on this?

With some part of this new implementation I managed to replace all of curve25519/448:
[ocheron:ed448](https://github.com/haskell-crypto/cryptonite/compare/master...ocheron:ed448)

This is interesting because it gives all EdDSA variants that are currently missing.

ocheron:
Additional notes:
- this implementation works correctly on a recent macOS machine but currently I have trouble building on Linux (where CC option `-mbmi2` is required)
- it is possible to replace only ed448 and keep current curve25519/ed25519 implementations, this would cut the code size roughly by half
- the library contains code for SHA-512 and SHAKE, ideally this should be more integrated with existing code in cryptonite
- to build with Cabal I had to copy some header files at several locations
- tested only on x86-64 so far, though I included also arch_32

ocheron:
Compilation issues with gcc should be fixed now (thanks to a change upstream).
I used cacophony benchmarks to see how implementations perform.
The improvement for 25519 is modest, but indeed is really good for 448.

Current code:
```
benchmarking Noise_NN_25519_ChaChaPoly_SHA256
time                 297.2 μs   (296.9 μs .. 297.6 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 297.3 μs   (296.9 μs .. 297.6 μs)
std dev              1.116 μs   (912.5 ns .. 1.384 μs)

benchmarking Noise_NN_448_ChaChaPoly_SHA256
time                 2.047 ms   (2.043 ms .. 2.050 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 2.053 ms   (2.049 ms .. 2.057 ms)
std dev              12.51 μs   (10.10 μs .. 15.81 μs)
```

New code:
```
benchmarking Noise_NN_25519_ChaChaPoly_SHA256
[...]

PR #167, opened 2017-06-05
Improve Curve448 and add Ed448

ocheron:
This is a new attempt at #84, keeping only Decaf reference implementations and _not_ changing 25519.

First I'd like to know if this simpler approach with no Cabal flag nor assembly looks safe enough.
In my tests I have recent Linux, macOS, Windows, OpenBSD working correctly.

Then there is the question if we consider upstream project stable enough.
To me this does not seem less mature than previous x448 implementation.


cbits/decaf/tools/generate.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/sh
+
+# Usage: ./generate.sh /path/to/ed448goldilocks-code
+#
+# Generate all files from ed448goldilocks branch 'master'
+# (available at <git://git.code.sf.net/p/ed448goldilocks/code>).
+#
+# Project is synced with upstream commit
+# '0a6e96827595fa1a5a62d12ac83c3cc5dda6dd67', i.e. tag 'v0.9.2'.
```
vincenthz, line 9: :1st_place_medal:

vincenthz:
that's looking good ! thanks @ocheron, That's a lot of painful work !

ocheron:
But much easier than writing the library...

I may have been too optimistic about the stability.  The version I took has a bug identified and fixed last week.  With possible bad consequences, so I will push an additional commit to master.

I still think this is the way forward: otherwise no bug is ever found.

PR #168, opened 2017-06-10
Validate P256 point when decoding

ocheron:
Adds validation of coordinates that is missing in P256 `decodePoint` implementation, and the beginning of a test module for `Crypto.ECC`.

### DIFF
--- a/Crypto/ECC.hs
+++ b/Crypto/ECC.hs
@@ -7,6 +7,7 @@
 --
 -- Elliptic Curve Cryptography
 --
+{-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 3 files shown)


Crypto/ECC.hs
```diff
@@ -114,7 +118,7 @@ instance EllipticCurve Curve_P256R1 where
         Nothing -> CryptoFailed $ CryptoError_PointSizeInvalid
         Just (m,xy)
             -- uncompressed
-            | m == 4 -> P256.pointFromBinary xy
+            | m == 4 -> P256.pointFromBinary xy >>= validateP256Point
```
vincenthz, line 121: Might be better to move to `pointFromBinary` being the same function that does the validation internally and maybe expose `unsafePointFromBinary` which does what `pointFromBinary` is doing now.

vincenthz (approved):

vincenthz:
thanks @ocheron , nice work !

Crypto/ECC.hs
ocheron: OK I'll look at it. Also in Haddock comments we could have more details for some functions.

PR #173, opened 2017-06-19
Remove support_blake2_sse flag in favor of support_sse

centromere:
The description of the `support_sse` flag said, "Use SSE optimized version when existing (BLAKE2, ARGON2)", so I figured that at some point someone wanted to have this done.


ocheron:
Yes we don't need two distinct flags.

PR #175, opened 2017-06-25
Fix many cases of unaligned accesses

jrtc27:
This is a second attempt at #110 which I have finally had a chance to do. The AES block128 changes feel slightly unsatisfactory, but copying to temporary buffers would be far more invasive, given the number of different variations, and my guess is that most of the operations are acting on either the input or output and thus could be unaligned. Since on x86 need_alignment is the constant 0, the compiler should be able to optimise away the if statement and with sufficient inlining be as efficient as before, but I haven't actually checked this. If you would prefer to also add an extra `#ifdef UNALIGNED_ACCESS_OK` check to define the block128 functions to their aligned versions I can do that, but that's more code, and I hope it's not needed.

### DIFF
--- a/cbits/aes/block128.h
+++ b/cbits/aes/block128.h
@@ -32,6 +32,7 @@
 #define BLOCK128_H
 
 #include <cryptonite_bitfn.h>
+#include <cryptonite_align.h>
 
 typedef union {
        uint64_t q[2];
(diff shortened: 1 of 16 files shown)


ocheron:
I don't know AES code well enough to be sure, but my intuition is that AES alignment issues come only from specific locations. So I would prefer a default function with `block128 *` arguments like block128_copy not to deal with alignment, and use other pointer types to mark buffers as possibly unaligned (like block128_copy_bytes).

IIRC freshly-allocated ByteArray instances are aligned to 8 bytes or more, so most output buffers won't be an issue for example. Have you identified where failures are located?

vincenthz:
I agree with @ocheron, but I think it's almost generally better than what we currently have, so will merge. 

It would be nice to have the alignment requirements in the types and the code (alignment assert) and check through the alignment check flag throughly, so we can catch potential alignment problem with travis.

PR #176, opened 2017-06-26
clarify padding requirements for PKCS15 encrypt/decrypt

frasertweedale:
The types do not say whether it is necessary to apply pad/unpad to
the input/output of the PKCS15 encrypt/decrypt functions.  Add
comments to clarify that it is not necessary to manually pad/unpad
the message.

### DIFF
--- a/Crypto/PubKey/RSA/PKCS15.hs
+++ b/Crypto/PubKey/RSA/PKCS15.hs
@@ -137,6 +137,8 @@ unpad packed
 -- information from the timing of the operation, the blinder can be set to None.
 --
 -- If unsure always set a blinder or use decryptSafer
+--
+-- The message is returned un-padded.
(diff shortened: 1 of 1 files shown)


Crypto/PubKey/RSA/PKCS15.hs
```diff
@@ -156,7 +158,8 @@ decryptSafer pk b = do
 
 -- | encrypt a bytestring using the public key.
 --
--- the message needs to be smaller than the key size - 11
+-- The message needs to be smaller than the key size - 11.
+-- The message should not padded.
```
tekul: I think you missed a "be" in there :).

frasertweedale: Not the bees!  (Thanks :D )

Issue #178, opened 2017-06-27
ECDH and ECIES failures


ocheron:
ECIES is supposed to check that ECDH result is not the point at infinity but this is not currently possible from the type class EllipticCurveDH.

X25519 has cofactor 8 and some (rare) weak points result in zeros when multiplied by any scalar.
This is because scalars are divisible by 8 and the all-zero result encodes both ∞ and (0,0), which has order 2.

For example:
```
> :m Crypto.ECC Crypto.Error Data.Proxy Data.ByteArray Crypto.PubKey.Curve25519
> let curve = Proxy :: Proxy Curve_X25519
> s <- curveGenerateScalar curve
> let CryptoPassed p = publicKey (pack [0x5f,0x9c,0x95,0xbc,0xa3,0x50,0x8c,0x24,0xb1,0xd0,0xb1,0x55,0x9c,0x83,0xef,0x5b,0x04,0x44,0x5c,0xc4,0x58,0x1c,0x8e,0x86,0xd8,0x22,0x4e,0xdd,0xd0,0x9f,0x11,0x57] :: Bytes)
> convert (ecdh curve s p) :: Bytes
"\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL"
```

To detect the situation one option would need to change EllipticCurveDH and ECIES functions with CryptoFailable results.

Another option is to raise an exception.
Some implementations of `ecdh` already fail when the scalar is zero:
```
> :m Crypto.ECC Data.Proxy Data.ByteArray
> let curve = Proxy :: Proxy Curve_P384R1
> let s = read "Scalar 0" :: Scalar Curve_P384R1
> p <- keypairGetPublic `fmap` curveGenerateKeyPair curve
> convert (ecdh curve s p) :: Bytes
"*** Exception: Crypto/ECC.hs:152:9-44: Irrefutable pattern failed for pattern Point x _
```
(but an exception from the private scalar is much less an issue than from the public point).

vincenthz:
I think this is one of those things where it's not a problem to break API, if it's border line a security problem or something that is painful to deal with

ocheron:
OK great, so this is also a change to include before releasing tls HEAD.

vincenthz:
[...]

PR #179, opened 2017-07-05
Validate result of P256.pointFromBinary and EllipticCurveDH.ecdh

ocheron:
This PR:
- changes API so that class EllipticCurveDH and module ECIES can return failures (stricter validation explained in #178)
- adds point validation directly in P256.pointFromBinary (discussed in #168)

### DIFF
--- a/Crypto/PubKey/ECIES.hs
+++ b/Crypto/PubKey/ECIES.hs
@@ -25,6 +25,7 @@ module Crypto.PubKey.ECIES
     ) where
 
 import           Crypto.ECC
+import           Crypto.Error
 import           Crypto.Random
 import           Crypto.Internal.Proxy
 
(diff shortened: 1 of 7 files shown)


Crypto/PubKey/ECIES.hs
```diff
@@ -33,16 +34,16 @@ import           Crypto.Internal.Proxy
 deriveEncrypt :: (MonadRandom randomly, EllipticCurveDH curve)
               => proxy curve -- ^ representation of the curve
               -> Point curve -- ^ the public key of the receiver
-              -> randomly (Point curve, SharedSecret)
+              -> randomly (CryptoFailable (Point curve, SharedSecret))
 deriveEncrypt proxy pub = do
     (KeyPair rPoint rScalar) <- curveGenerateKeyPair proxy
-    return (rPoint, ecdh proxy rScalar pub)
+    return $ (\s -> (rPoint, s)) `fmap` ecdh proxy rScalar pub
```
vincenthz, line 40: it would probably make more sense to retry here instead of failing, since we control the scalar completely. it seems unlikely to happens either, so we could also ignore it, but retrying seems like the best thing to do here.

ocheron: I thought about it too, but went back: if the receiver public value `pub` is in a small subgroup, the multiplication fails for every scalar.

vincenthz: should point validation make sure that doesn't happens ?

vincenthz: ok realized this doesn't make sense generally.

vincenthz:
Thanks !

PR #189, opened 2017-09-17
CAST5 block cipher

### LINKED ISSUE #33: Symmetric Ciphers: Implement missing ones
- [ ] CAST5 / CAST128 (RFC 2144)
- [ ] Twofish (https://www.schneier.com/paper-twofish-paper.pdf)

ocheron:
A direct implementation following RFC 2144 with everything unboxed.
This should allow to close #33.


vincenthz:
thanks ! Maybe I've read too many cipher, but that looks quite readable !

ocheron:
Well this one is not too complex and was a perfect example to give block ciphers a try.

FYI I had also looked at `Block Word32` from foundation as buffer backend but didn't send this as you just reorganized things with basement. Foundation reexports unsafe read/write functions with `MutableCollection` but I didn't find functions like `index` and `unsafeIndex`. Maybe that is just missing in `IndexedCollection` ?

PR #191, opened 2017-09-19
prefer /dev/urandom over /dev/random

piyush-kurur:
On Unix machines it is in general preferable to use `/dev/urandom` instead of `/dev/random`. If I understand the code correctly this one line patch will do the necessary.

### DIFF
--- a/Crypto/Random/Entropy/Backend.hs
+++ b/Crypto/Random/Entropy/Backend.hs
@@ -35,7 +35,7 @@ supportedBackends =
 #ifdef WINDOWS
     openBackend (undefined :: WinCryptoAPI)
 #else
-    openBackend (undefined :: DevRandom), openBackend (undefined :: DevURandom)
+    openBackend (undefined :: DevURandom), openBackend (undefined :: DevRandom)
 #endif
     ]
 


centromere:
Why is this preferred? My understanding is that `/dev/random` has better entropy than `/dev/urandom`.

piyush-kurur:
https://www.2uo.de/myths-about-urandom/

piyush-kurur:
There are even better alternatives now. On linux (modern kernel with appropriate libc) there is the getrandom call and on OpenBSD/NetBSD there is the arc4random calls which are significantly better alternatives.

piyush-kurur:
Refer to #20 for a problem with blocking `/dev/random`. I would go ahead and even suggest solely depending on `/dev/urandom` for all unixes and ignore the `/dev/random`.

vincenthz:
none of the entropy sources are supposed to block with cryptonite, it was supposed to work this way and got fixed in #20 when it didn't. While I agree that urandom is enough for linux and BSD, I don't see anything harmful in trying to source random first and it could be useful for (unsupported) systems where `/dev/random` is actually better than `/dev/urandom`. If we'ld move `/dev/urandom` to be first source of entropy, then it's very likely that `/dev/random` would just never be used.

Also, I'll very likely do a PR to remove (or complement) the entropy gathering in cryptonite and use foundation's one which have binding to the linux syscall `getrandom`.

piyush-kurur:
`getrandom` is better on linux. It is available since Kernel-3.17. However, remember that the man page on linux is rather misleading. While the kernel does implement the getrandom syscall, glibc did not have support for a long time (Only > 2.25). Even debian testing does not have this. So you will need to hack around a bit.

vincenthz:
@piyush-kurur it's already implemented and working in https://github.com/haskell-foundation/foundation/blob/master/cbits/foundation_random.c#L22-L47 (it's not using glibc)

piyush-kurur:
Good to know thanks.

piyush-kurur:
If the use of `getrandom` from foundation is to be merged soon then I suggest that you close this ticket. Alternatively since this is a minor 1-line change you can merge it.

What is the situation with other platform specific things like arc4random on OpenBSD/NetBSD systems ?

vincenthz:
related to the OpenBSD/NetBSD, foundation doesn't have specific support (i.e. it's using the standard unix /dev/(u)random stuff), but there's no reason where it can't be added there if someone is motivated by this.

piyush-kurur:
On Mon, Oct 02, 2017 at 02:12:01PM -0700, Vincent Hanquez wrote:
> related to the OpenBSD/NetBSD, foundation doesn't have specific support (i.e. it's using the standard unix /dev/(u)random stuff), but there's no reason where it can't be added there if someone is motivated by this.

Let me see if I can send you a pull request for that. I have a version
of it in raaz which I have checked on an openbsd machine that I had
which sadly I do not have as of now.

From what you have said so far, it would be better to contribute
it to foundation rather than.

mithrandi:
> While I agree that urandom is enough for linux and BSD, I don't see anything harmful in trying to source random first and it could be useful for (unsupported) systems where /dev/random is actually better than /dev/urandom.

I'm not aware of any systems on which /dev/random and /dev/urandom have separate entropy sources; on some platforms they are identical (eg. BSD), whereas on Linux reading from /dev/random may block, but the source of the entropy is the same. I found this PR while trying to understand the current behaviour of the code; reading from both seems very confusing.

vincenthz:
As a random example, Some Solaris (that we don't really support) will use any plugged hardware RNG (if any) when using `/dev/random` but not `/dev/urandom`.

It's not reading from all source of entropy at once, it's trying one source after another to fill the demanded buffer size; if `/dev/random` happens to block, it will switch to the next source (being `/dev/urandom`) when we get a `EAGAIN`/`ENONBLOCK` (or any other errors).

vincenthz:
archiving repository

PR #199, opened 2017-10-23
Use Proxy for openBackend

ocheron:
Project still works with GHC 7.6 but with the Proxy compatibility module it's easy to replace `undefined`.

We can also bump `base` to 4.6 minimum after #182.


vincenthz:
Thanks !

michalrus:
Thank you!

PR #202, opened 2017-11-20
Ed25519 arithmetic primitives

ocheron:
This adds a new module `Crypto.ECC.Ed25519` with the usual point and scalar primitives implemented as wrappers of ed25519-donna code. The main part is arbitrary scalar multiplication which does not exist in the original code but added locally.

One caveat in ed25519-donna is that point decoding is not time constant because the square-root code tries 1 or 2 possible results. But that seems minor.

The PR also extends class EllipticCurveArith with point negation, adding P256 implementation that did not exist, and changing the ECC implementation so that coordinates stay in range [0..p)


vincenthz:
I've started to review this twice now and got interrupted every time. I'll make sure this get review sometimes this week, sorry about the delay, it looks like an amazing PR albeit slightly difficult related to ed25519 specificity.

ocheron:
Thanks @vincenthz just take as much time as you need.

Actually I'll be very happy to know if @jml or you (or anyone else) have some use for this and what are the results.

I obviously took inspiration from preexisting techniques and code. There are not many comments and I can try to fill the gaps where needed.

jml:
@ocheron I would definitely update [spake2](http://hackage.haskell.org/package/spake2) to use this (ugh, the docs that I worked so hard on are not online). I'm not able try it out before it gets merged though—sorry.

tests/ECC/Ed25519.hs
```diff
@@ -0,0 +1,119 @@
...
+            let pR = Ed25519.pointMul x p `Ed25519.pointAdd` Ed25519.pointMul y p
+             in pR `propertyEq` Ed25519.pointMul (x `Ed25519.scalarAdd` y) p
+        , testProperty "double scalarmult" $ \n1 n2 p ->
+            let pR = Ed25519.pointAdd (Ed25519.toPoint n1) (Ed25519.pointMul n2 p)
+             in pR `propertyEq` Ed25519.pointsMulVarTime n1 n2 p
+        ]
+    ]
+  where
+    p0 = Ed25519.toPoint s0
+    CryptoPassed s0 = Ed25519.scalarDecodeLong ("" :: ByteString)
```
vincenthz: it might be more informative to replace a pattern match by a `throwCryptoError`. although I suspect we'ld be losing the location error right now by doing so, so might be adding a CallStack context to `throwCryptoError` which would be an improvement by itself anyway ?

Crypto/PubKey/ECC/Prim.hs
```diff
@@ -30,9 +31,9 @@ scalarGenerate curve = generateBetween 1 (n - 1)
 -- | Elliptic Curve point negation:
 -- @pointNegate c p@ returns point @q@ such that @pointAdd c p q == PointO@.
 pointNegate :: Curve -> Point -> Point
-pointNegate _           PointO     = PointO
-pointNegate CurveFP{}  (Point x y) = Point x (-y)
-pointNegate CurveF2m{} (Point x y) = Point x (x `addF2m` y)
+pointNegate _            PointO     = PointO
+pointNegate (CurveFP c) (Point x y) = Point x (ecc_p c - y)
```
vincenthz, line 35: that looks like it's fixing a serious bug in pointNegate, or changing the sign of y eventually "even out" later in the modular operation ?

Crypto/ECC/Simple/Prim.hs
```diff
@@ -49,7 +50,7 @@ pointNegate :: Curve curve => Point curve -> Point curve
 pointNegate        PointO     = PointO
 pointNegate point@(Point x y) =
     case curveType point of
-        CurvePrime {}  -> Point x (-y)
+        CurvePrime (CurvePrimeParam p) -> Point x (p - y)
```
vincenthz, line 53: same remark as the other pointNegate call. I think it's clearly more accurate this way and keep y in the `Fe` range (maybe normalized would be the right term)

Crypto/ECC/Ed25519.hs
```diff
@@ -0,0 +1,298 @@
...
+    generate :: MonadRandom randomly => randomly ScrubbedBytes
+    generate = getRandomBytes 32
+
+    -- Uses the same bit mask than during key-generation procedure,
+    -- but without making divisible by 8.  As a consequence of modular
+    -- reduction, distribution is not uniform.  But the curve order is
+    -- very close to 2^252 so only a tiny fraction of the scalars have
+    -- lower probability, roughly 1/(2^126) of all possible values.
+    clamp :: ByteArrayAccess ba => ba -> ScrubbedBytes
+    clamp bs = B.copyAndFreeze bs $ \p -> do
```
vincenthz: I'm not sure about this; so there's no reason to make it divisible by 8 for ed25519, however with the birationally equivalence with curve25519 and curve25519 low order requirements (divisible by 8), I'm not sure how to best handle this in a non-confusing way or if we should at all.

The same thing with the rest of the clamping, the reason for it to be done is actually slightly obscure and not as specified as I would like (I spent a good amount of time of this very problem recently). The only reason that I found for this clamping is to actually prevent a timing issue with multiplication overflowing, but once you take away the specific usage to the signature system (ed25519), and instead do generic mathematical operation on `edwards25519` does it make sense to still have this ?

Crypto/ECC/Ed25519.hs
```diff
@@ -0,0 +1,298 @@
...
+        withByteArray p1 $ \pp1 ->
+        withByteArray p2 $ \pp2 ->
+            fmap (/= 0) (ed25519_point_eq pp1 pp2)
+    {-# NOINLINE (==) #-}
+
+-- | Generate a random scalar.
+scalarGenerate :: MonadRandom randomly => randomly Scalar
+scalarGenerate = unwrap . scalarDecodeLong . clamp <$> generate
+  where
+    unwrap (CryptoPassed x) = x
```
vincenthz: `unwrap` looks like `throwCryptoError`

Crypto/ECC/Ed25519.hs
```diff
@@ -0,0 +1,298 @@
+-- |
+-- Module      : Crypto.ECC.Ed25519
```
vincenthz: In some other related works, I tried to distinguish Ed25519 (the signature system) from the curve (Edwards25519). do you think it might be worthwhile to do that here at the module level ?

Crypto/ECC/Ed25519.hs
```diff
@@ -0,0 +1,298 @@
...
+-- Format is binary compatible with 'Crypto.PubKey.Curve25519.SecretKey'
+-- from module "Crypto.PubKey.Curve25519".
+scalarEncode :: B.ByteArray bs => Scalar -> bs
+scalarEncode (Scalar s) =
+    B.allocAndFreeze 32 $ \out ->
+        withByteArray s $ \ps -> ed25519_scalar_encode out ps
+
+-- | Deserialize a little-endian number as a scalar.  Input array can
+-- have any length from 0 to 64 bytes.
+scalarDecodeLong :: B.ByteArrayAccess bs => bs -> CryptoFailable Scalar
```
vincenthz: I'm not sure about this. The way I see it that the extended secret key of ed25519 is actually a two part type; one half is a 32 bytes scalar (in little endian) that is technically the secret value which derive the public key. While the second half is a prefix used during the signature process only.

at the ed25519 level, the 32 bytes secret key is hashed with SHA512 to make this 64 bytes extended quantity; I wonder if that would be a good idea to distinguish this here at the type level instead of considering it an opaque scalar.

cbits/ed25519/ed25519.c
```diff
@@ -11,6 +11,7 @@
 #include "ed25519.h"
 #include "ed25519-randombytes.h"
 #include "ed25519-hash.h"
+#include "ed25519-cryptonite-exts.h"
```
vincenthz, line 14: I think that's great to distinguish the exts parts. I did some extensions in another project, and should really have use this kind of idea !

vincenthz (approved):

Crypto/ECC/Ed25519.hs
ocheron: Yes I was entirely focused on the prefix `Crypto.ECC` vs `Crypto.PubKey` but renaming the module `Edwards25519` will make it even clearer. So I'll change it, as well as the curve name for consistency.

Crypto/ECC/Ed25519.hs
ocheron: OK I'll remove it.

Crypto/ECC/Ed25519.hs
ocheron: The masking of highest bits is something I tried to understand too and indeed is not really explained. In addition to having a fixed magnitude for timing reasons, I wanted to limit the number of times values wrap around the curve order and decrease the fraction of scalars with statistical bias. However I see libraries take another approach, actually using many more bits [like this](https://github.com/warner/python-spake2/blob/05b9f968d37dc5419f0e6e20c9b65737de21a7e9/src/spake2/ed25519_basic.py#L162) so I think I'm going to change that part.

The scalar functions work modulo prime order L, not 8*L, so masking the lower bits does not provide divisibility by 8. Here the scalars will be most useful when applied to points in the prime-order subgroup.

It's true all of this gives enough power to shoot oneself in the foot with the cofactor.
However I don't see how to keep generality to implement many scheme and at the same time prevent any effect of cofactor (besides the Decaf technique of switching to another curve).

But clearly this deserves some warnings in documentation.

Crypto/ECC/Ed25519.hs
ocheron: Yes ideally decoding should accept a scalar of any length and never fail. In libdecaf this is like this. In ed25519-donna, the code uses `unsigned char work[64]` as temporary buffer, so here I'm just exposing and documenting the limitation in Haskell as well.

Crypto/PubKey/ECC/Prim.hs
ocheron: I found the bug when writing P256 tests and comparing results with ECC.
But so far this has not been serious: `pointNegate` is not exposed, only used internally in a way that does not matter.

cbits/ed25519/ed25519.c
ocheron: Almost everything is in header files, so I'm just continuing this trend and working around static declarations...

tests/ECC/Ed25519.hs
ocheron: I'll look into it.

vincenthz (approved):
nice documentation addition !

ocheron:
I think I finished it all, except the CallStack part which is probably best in a separate PR and after #206 (allowing to use `Basement.Compat.CallStack`).

The tests are extended to all curve points and I added explanation and functions to help prevent mistakes because of the cofactor.

vincenthz:
You're reading my mind on CallStack :+1: No reason to wait more for merging #206 I think.

vincenthz:
great work on edwards25519 !

jml:
Indeed! Thank you both so much.

@vincenthz, any chance you can cut a release of cryptonite for this change?

ocheron:
Now available in cryptonite-0.25 if you want to try @jml.

Note for those interested in Ed448-Goldilocks: we don't have an exact implementation of Edwards448 in libdecaf but one of a 4-isogenous curve instead. This means it is not possible to define the `EllipticCurveArith` instance and simultaneously keep a basic expectation like `pointDecode.pointEncode==id`.

Still looking what kind of alternate API would be best in this situation, without tying too much to the decaf implementation if possible.

PR #205, opened 2017-11-30
withDev: Close the FD on exception

osa1:
Currently I think it's very _unlikely_ but _possible_ to have fd leaks: `withDev`
is called with two arguments: `gatherDevEntropy` and
`gatherDevEntropyNonBlock`. Both handle `IOException`s, but it's possible that
the thread gets an async exception, which is not handled and causes `withDev` to
leak the fd.

With this PR `withDev` closes the fd in all cases.


vincenthz:
Thanks, indeed probably better this way even though it's unlikely

Issue #207, opened 2017-11-30
Build failure on GHC 7.8.4 with foundation-0.0.13


RyanGlScott:
As observed in [this Travis build](https://travis-ci.org/scotty-web/wai-middleware-static/jobs/309686391#L784), `cryptonite-0.24` fails to build on GHC 7.8.4 against `foundation-0.0.13`:

```
[ 37 of 120] Compiling Crypto.Hash.Types ( Crypto/Hash/Types.hs, dist/build/Crypto/Hash/Types.o )

Crypto/Hash/Types.hs:66:22:
    No instance for (ByteArrayAccess (F.UArray Word8))
      arising from the 'deriving' clause of a data type declaration
    Possible fix:
      use a standalone 'deriving instance' declaration,
        so you can specify the instance context yourself
    When deriving the instance for (ByteArrayAccess (Digest a))
```

`foundation-0.0.13` appears to be involved, since `cryptonite-0.24` _does_ build correctly on GHC 7.8.4 against `foundation-0.0.15`.

vincenthz:
I haven't foreseen this, but that make sense. memory-0.14.9 which I released yesterday now depends on basement's implementation, and you need at least foundation 0.0.14 to have basement. so that leave foundation-0.0.13 with it's own F.UArray type (which is not basement's), so no instances.

I'm not sure if there's anything that can be done to express this state of things (also whether it's worth it, 7.8 is getting long in the tooth).

somewhat related to #206

RyanGlScott:
OK. Can you simply adjust the lower version bounds on `foundation` to require 0.0.14 or later?

phadej:
Observable in https://github.com/haskell-servant/servant/pull/868 and travis job: https://travis-ci.org/haskell-servant/servant/jobs/310251850

phadej:
The trickiness is that https://matrix.hackage.haskell.org/package/cryptonite supports even 7.6, when latest foundation only 7.8: https://matrix.hackage.haskell.org/package/foundation

Setting `foundation >= 0.14` would effectively drop GHC 7.6 support.

*EDIT:* though 7.6 isn't even tested in travis matrix, so seems to compile (hopefully also work!) by accident.

hvr:
This is messy... what's the conclusion/summary? 

[...]

PR #208, opened 2017-12-05
Foundation flag

phadej:
Disallows invalid combination of `foundation <0.14, memory >=0.14.9`, see https://github.com/haskell-crypto/cryptonite/issues/207 

Simpler approach would be to bump the `foundation` lower bound to `foundation >=0.14`, but that will disallow (non-officially supported) GHC-7.6.3 install plans.

### DIFF
--- a/cryptonite.cabal
+++ b/cryptonite.cabal
@@ -102,6 +102,11 @@ Flag check_alignment
   Default:           False
   Manual:            True
 
+Flag foundation-0_14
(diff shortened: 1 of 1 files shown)


phadej:
Travis failure (GHC-8.0.2) is due slow Travis, could some one restart the job?

vincenthz:
that might be revived later when we start dropping old versions but for now I'll close this

PR #210, opened 2018-01-09
More efficient digestFromByteString

avieth:
By avoiding `ByteArray.convert`, `digestFromByteString` is much more efficient. 391 bytes allocated versus 1917. 135 nanoseconds versus 1.286 microseconds.

PR #211, opened 2018-01-09
byteStringFromDigest defined

avieth:
AFAIK the standard way to get a `ByteString` from a `Digest h` is to use `ByteArray.convert`, but that's not so efficient. `byteStringFromDigest` is defined here, in 3 different ways. The specialized `byteStringFromDigest' :: Digest h -> ByteString` is fastest, but we believe it should be possible to get a `byteStringFromDigest'' :: ByteArray ba => Digest h -> ba` which is just as good.
  
Follow up to #210

### DIFF
--- a/Crypto/Hash.hs
+++ b/Crypto/Hash.hs
@@ -26,6 +26,7 @@ module Crypto.Hash
     -- * Functions
     , digestFromByteString
     , byteStringFromDigest
+    , convertByteStringFromDigest
     -- * Hash methods parametrized by algorithm
     , hashInitWith
     , hashWith
@@ -51,7 +52,7 @@ import           Control.Monad
 import           Crypto.Internal.Compat (unsafeDoIO)
 import           Crypto.Hash.Types
 import           Crypto.Hash.Algorithms
-import           Foreign.Ptr (Ptr)
+import           Foreign.Ptr (Ptr, castPtr)
 import           Crypto.Internal.ByteArray (ByteArrayAccess, ByteArray)
 import qualified Crypto.Internal.ByteArray as B
 import           Data.ByteString (ByteString)
@@ -129,9 +130,17 @@ digestFromByteString = from undefined
             count = CountOf (B.length ba)
 {-# INLINABLE digestFromByteString #-}
 
--- | Get the bytes of a Digest.
 byteStringFromDigest :: forall a ba . (ByteArray ba) => Digest a -> ba
-byteStringFromDigest (Digest uarray@(UArray _ (CountOf size) _)) = unsafeDoIO $ do
-    (_, ba) <- B.allocRet size (B.copyByteArrayToPtr uarray)
+byteStringFromDigest digest = unsafeDoIO $ do
+    (_, ba) <- B.allocRet (B.length digest) (B.copyByteArrayToPtr digest)
     return ba
 {-# INLINABLE byteStringFromDigest #-}
+
+convertByteStringFromDigest :: forall a ba . (ByteArray ba) => Digest a -> ba
+convertByteStringFromDigest digest = unsafeDoIO (go (B.length digest))
+  where
+    go sz | sz < 0 = go 0
+          | otherwise = fmap snd $ B.allocRet sz $ \d -> do
+                            B.copyByteArrayToPtr digest d
+                            (\_ -> return ()) (castPtr d)
+{-# INLINABLE convertByteStringFromDigest #-}
--- a/Crypto/Hash.hs
+++ b/Crypto/Hash.hs
@@ -25,6 +25,9 @@ module Crypto.Hash
     , Digest
     -- * Functions
     , digestFromByteString
+    , byteStringFromDigest
+    , byteStringFromDigest'
+    , byteStringFromDigest''
     -- * Hash methods parametrized by algorithm
     , hashInitWith
     , hashWith
@@ -43,14 +46,18 @@ module Crypto.Hash
 
 import           Basement.Types.OffsetSize (CountOf (..))
 import           Basement.Block (Block, unsafeFreeze)
-import           Basement.Block.Mutable (copyFromPtr, new)
+import           Basement.Block.Mutable (copyFromPtr, copyToPtr, new)
+import           Basement.UArray (UArray (..))
+import qualified Basement.UArray as UA (copyToPtr)
 import           Control.Monad
 import           Crypto.Internal.Compat (unsafeDoIO)
 import           Crypto.Hash.Types
 import           Crypto.Hash.Algorithms
 import           Foreign.Ptr (Ptr)
-import           Crypto.Internal.ByteArray (ByteArrayAccess)
+import           Crypto.Internal.ByteArray (ByteArrayAccess, ByteArray)
 import qualified Crypto.Internal.ByteArray as B
+import           Data.ByteString (ByteString)
+import qualified Data.ByteString.Internal as BS
 import qualified Data.ByteString.Lazy as L
 import           Data.Word (Word8)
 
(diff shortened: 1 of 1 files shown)


vincenthz:
I'm thinking that solving this problem at the `memory` level will be better. it just need a new method to do the "copyToPtr" step.

avieth:
> I'm thinking that solving this problem at the memory level will be better. it just need a new method to do the "copyToPtr" step.

A new method on the `ByteArray` class?

vincenthz:
it can go directly in `ByteArrayAccess`, since it's a readonly access and we're not creating a bytearray.

The default method is basically something like:

```haskell
method a dstptr = withByteArray a $ \srcptr -> memCopy dstptr srcptr (length a)
```

vincenthz:
actually I'm not sure anymore why I think it would help; Could you re-add the profile of each version here ?

avieth:
> actually I'm not sure anymore why I think it would help; Could you re-add the profile of each version here ?

First is `byteStringFromDigest`, second is `byteStringFromDigest'`, third is `byteStringFromDigest''`

```
benchmarking digest to bytestring/old
time                 249.6 ns   (248.5 ns .. 251.0 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 248.2 ns   (247.6 ns .. 249.6 ns)
std dev              2.776 ns   (1.402 ns .. 4.390 ns)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              631.552    (631.325 .. 631.764)
  y                  -324812.566 (-367873.209 .. -281971.764)

benchmarking digest to bytestring/new_1
time                 45.94 ns   (45.89 ns .. 46.00 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 45.81 ns   (45.75 ns .. 45.92 ns)
std dev              257.7 ps   (150.7 ps .. 454.1 ps)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              191.918    (191.867 .. 191.964)
  y                  -303415.946 (-342161.197 .. -265346.800)

benchmarking digest to bytestring/new_2
time                 76.16 ns   (76.09 ns .. 76.26 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 75.94 ns   (75.79 ns .. 76.32 ns)
std dev              705.3 ps   (348.3 ps .. 1.342 ns)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              279.863    (279.784 .. 279.924)
  y                  -322650.378 (-364123.753 .. -285487.324)
```

avieth:
The dubious y intercepts in the above snippet are due to a bug in criterion. I patched it and ran again.

```
benchmarking digest to bytestring/old
time                 247.1 ns   (246.8 ns .. 247.5 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 246.7 ns   (246.2 ns .. 248.4 ns)
std dev              2.821 ns   (802.5 ps .. 5.746 ns)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              632.000    (631.999 .. 632.002)
  y                  1026.428   (790.187 .. 1242.002)
variance introduced by outliers: 10% (moderately inflated)

benchmarking digest to bytestring/new_1
time                 44.47 ns   (44.44 ns .. 44.51 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 44.42 ns   (44.39 ns .. 44.48 ns)
std dev              141.8 ps   (98.83 ps .. 245.1 ps)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              192.000    (192.000 .. 192.000)
  y                  1127.706   (931.541 .. 1319.275)

benchmarking digest to bytestring/new_2
time                 77.47 ns   (77.37 ns .. 77.59 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 77.17 ns   (77.05 ns .. 77.29 ns)
std dev              397.8 ps   (337.6 ps .. 509.1 ps)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              280.000    (280.000 .. 280.001)
  y                  1096.868   (899.064 .. 1315.922)


```

vincenthz:
so this https://github.com/vincenthz/hs-memory/pull/43 + basement 0.0.5 should help with `convert` (and other use of `ByteArray` `allocRet`). There's still an extra copy which would be save by the copyToPtr method for Block and UArray.

avieth:
> so this vincenthz/hs-memory#43 + basement 0.0.5 should help with convert (and other use of ByteArray allocRet). There's still an extra copy which would be save by the copyToPtr method for Block and UArray.

@vincenthz Nice. I built against basement-0.0.5 and hs-memory on that branch and ran the benchmarks again. I also made each candidate INLINABLE.

Order of benchmarks is: `ByteArray.convert`, direct via `ByteString.Internal.create`, and the general `ByteArray` version.

Here are the results with `-O0`:

```
benchmarking digest to bytestring/old
time                 110.8 ns   (110.3 ns .. 111.5 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 110.5 ns   (110.1 ns .. 110.9 ns)
std dev              1.382 ns   (1.087 ns .. 1.775 ns)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              320.000    (319.999 .. 320.000)
  y                  2192.478   (1979.965 .. 2418.397)
variance introduced by outliers: 13% (moderately inflated)

benchmarking digest to bytestring/new_1
time                 43.09 ns   (42.96 ns .. 43.31 ns)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 42.92 ns   (42.75 ns .. 43.53 ns)
std dev              942.9 ps   (310.7 ps .. 1.909 ns)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              192.000    (192.000 .. 192.000)
  y                  2243.252   (2047.952 .. 2448.541)
variance introduced by outliers: 33% (moderately inflated)

benchmarking digest to bytestring/new_2
time                 74.23 ns   (74.05 ns .. 74.48 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 73.60 ns   (73.40 ns .. 73.88 ns)
std dev              783.2 ps   (560.6 ps .. 1.237 ns)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              280.000    (279.999 .. 280.000)
  y                  2278.788   (2059.558 .. 2506.066)
```

And here they are with `-O2`:

```
benchmarking digest to bytestring/old
time                 81.61 ns   (81.51 ns .. 81.72 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 80.99 ns   (80.82 ns .. 81.17 ns)
std dev              578.4 ps   (480.5 ps .. 721.4 ps)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              272.000    (272.000 .. 272.000)
  y                  2182.401   (1973.496 .. 2409.474)

benchmarking digest to bytestring/new_1
time                 31.26 ns   (31.22 ns .. 31.32 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 31.17 ns   (31.07 ns .. 31.41 ns)
std dev              493.3 ps   (207.7 ps .. 941.9 ps)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              152.000    (152.000 .. 152.000)
  y                  2289.044   (2076.980 .. 2503.107)
variance introduced by outliers: 20% (moderately inflated)

benchmarking digest to bytestring/new_2
time                 28.92 ns   (28.84 ns .. 29.01 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 28.71 ns   (28.63 ns .. 28.77 ns)
std dev              225.5 ps   (176.2 ps .. 293.6 ps)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              144.000    (144.000 .. 144.000)
  y                  2222.861   (2030.145 .. 2433.299)
```

With optimizations, general `ByteArray` version outperforms the direct `ByteString` version!

vincenthz:
I'm a bit puzzled why convert is not as fast as the latest one considering it's the same code. could it be that you'ld want a SPECIALIZE instead of INLINEABLE on the benchmarked function to make sure it's not benchmarking the generic Bytearray one ?

avieth:
> I'm a bit puzzled why convert is not as fast as the latest one considering it's the same code. could it be that you'ld want a SPECIALIZE instead of INLINEABLE on the benchmarked function to make sure it's not benchmarking the generic Bytearray one ?

Yeah I'm puzzled too. I've rebased onto master and benchmarked again. Using convert is about 85ns versus 30ns going direct. I tried specializing `allocRet` for `ByteString` and `copyByteArrayToPtr` for `Block Word8` but GHC always ignores it, claiming `memory` wasn't compiled with `-O` (it was) or that there's no inlinable pragma (there is).

avieth:
Changing `copyAndFreeze` from `NOINLINE` to `INLINABLE` makes `convert` significantly faster. @vincenthz is this a bad thing to do?

avieth:
@vincenthz Never mind my earlier 2 comments. I hadn't built against the latest `memory`! Things are good now: `convert` is only a couple nanoseconds slower than the direct definition.

avieth:
No action required. Improvements to upstream libraries have made performance better.

PR #219, opened 2018-02-11
Add Read instance for Digest type

ocheron:
Follows @paolino's idea of strong splitting in #78 with some modifications to eliminate intermediate allocations:
- using basement directly
- conversion from hexadecimal is pushed into splitting recursion


vincenthz:
nice :+1:

PR #222, opened 2018-03-12
Add an example of using Crypto.OTP for Google Authenticator

MaxGabriel:
Hi, I recently added 2FA to our site and thought this example code would help future users of cryptonite. There are some things that are easy to get wrong (notably: it's not clear that you can and should generate >80 bit keys, and what the formatting of the secret code should be). The sample code also points to the `memory` implementation of base32 encoding, which is nice because the other implementations on Hackage are unmaintained and/or buggy.

Here's how it looks rendered to Haddocks:

image

If you think these docs are too long, I could cut it back to just the code snippets to reduce length, or put the documentation in a Haddock collapsible section.

### DIFF
--- a/Crypto/OTP.hs
+++ b/Crypto/OTP.hs
@@ -26,6 +26,43 @@
 -- >>>
 -- >>> let getOTPTime = getPOSIXTime >>= \t -> return (floor t :: OTPTime)
 --
+-- == Example Usage: Integration with Google Authenticator
+--
+-- Google Authenticator imposes a few restrictions on TOTP, such as only supporting 6 digit codes and 30 second time windows. An implementation would look something like:
+--
+-- 1. Generate a secret key and store it in a database, marking the code as unverified.
+-- 2. Return the encoded secret key to the client.
+-- 3. The client should render a QR code with the data: @otpauth:\/\/totp\/URL_ENCODED_COMPANY_NAME?secret=ENCODED_SECRET&issuer=URL_ENCODED_COMPANY_NAME@
+-- 4. The user should scan the QR code using Google Authenticator, then submit their 6 digit code ('OTP').
+-- 5. Use 'totpVerify' with 'defaultTOTPParams' to validate the code, then mark the code as verified.
+-- 6. Future login attempts should require a new 'OTP' code.
+--
+-- This code demonstrates how to generate the secret key:
+--
+-- > import Control.Monad.IO.Class
+-- > import qualified Data.Text as T
+-- > import qualified Data.Text.Encoding as TE
+-- > import Data.ByteArray (ByteArrayAccess) -- memory package
+-- > import Data.ByteArray.Encoding (convertToBase, Base(Base32))
+-- > import Crypto.Random (getSystemDRG, randomBytesGenerate)
+-- >
+-- > newtype TOTPSecretKey = TOTPSecretKey ByteString
+-- >     deriving (Show, ByteArrayAccess)
+-- > 
+-- > newtype GoogleAuthenticatorEncodedKey = GoogleAuthenticatorEncodedKey Text
+-- >   deriving (Show)
+-- >
+-- > -- Google Authenticator accepts arbitrary length secrets, using 80 bits for its own services.
+-- > -- RFC 4226 recommends a 160 bit (20 byte) secret key https://tools.ietf.org/html/rfc4226#section-4
+-- > randomTOTPSecretKey :: MonadIO m => m (TOTPSecretKey, GoogleAuthenticatorEncodedKey)
+-- > randomTOTPSecretKey = do
+-- >   key <- liftIO $ fst . randomBytesGenerate 20 <$> getSystemDRG
+-- >   return (TOTPSecretKey key, googleAuthenticatorEncode key)
+-- >
+-- > googleAuthenticatorEncode :: ByteString -> GoogleAuthenticatorEncodedKey
+-- > googleAuthenticatorEncode bs = GoogleAuthenticatorEncodedKey $ stripPadding $ TE.decodeUtf8 $ convertToBase Base32 bs
+-- >   where
+-- >     stripPadding t = T.dropWhileEnd (== '=') t
 
 module Crypto.OTP
     ( OTP


Crypto/OTP.hs
```diff
@@ -26,6 +26,43 @@
 -- >>>
 -- >>> let getOTPTime = getPOSIXTime >>= \t -> return (floor t :: OTPTime)
 --
+-- == Example Usage: Integration with Google Authenticator
+--
+-- Google Authenticator imposes a few restrictions on TOTP, such as only supporting 6 digit codes and 30 second time windows. An implementation would look something like:
+--
+-- 1. Generate a secret key and store it in a database, marking the code as unverified.
+-- 2. Return the encoded secret key to the client.
```
tekul, line 34: I think I'd be confused at this point about what the "the code" and "the encoded secret key" are. Are they they same thing? I'd mention beforehand in the summary that GA uses Base-32 encoding for the key. Maybe avoid the term "code" as it's kind of overloaded (also "QR code")? Perhaps use "passcode" when you're referring to the string of numbers entered by the user.

Also it's not clear what has to be marked as unverified to begin with and why.

Crypto/OTP.hs
```diff
@@ -26,6 +26,43 @@
...
+-- 3. The client should render a QR code with the data: @otpauth:\/\/totp\/URL_ENCODED_COMPANY_NAME?secret=ENCODED_SECRET&issuer=URL_ENCODED_COMPANY_NAME@
+-- 4. The user should scan the QR code using Google Authenticator, then submit their 6 digit code ('OTP').
```
tekul, line 36: It might be simpler to just say that the user has to enter the encoded secret key into GA, rather than going into QR codes here. Entering it manually is the simplest implementation and also probably more secure, since you can choose the label yourself rather than having it obviously linked to a particular company account.

Crypto/OTP.hs
```diff
@@ -26,6 +26,43 @@
...
+-- 5. Use 'totpVerify' with 'defaultTOTPParams' to validate the code, then mark the code as verified.
+-- 6. Future login attempts should require a new 'OTP' code.
```
tekul, line 38: Again, I'd want to know what "mark the code as verified" means. Also a line of (Haskell :-) ) code showing the OTP validation step would be nice.

Crypto/OTP.hs
```diff
@@ -26,6 +26,43 @@
...
+-- > import Data.ByteArray.Encoding (convertToBase, Base(Base32))
+-- > import Crypto.Random (getSystemDRG, randomBytesGenerate)
+-- >
+-- > newtype TOTPSecretKey = TOTPSecretKey ByteString
+-- >     deriving (Show, ByteArrayAccess)
+-- > 
+-- > newtype GoogleAuthenticatorEncodedKey = GoogleAuthenticatorEncodedKey Text
+-- >   deriving (Show)
+-- >
+-- > -- Google Authenticator accepts arbitrary length secrets, using 80 bits for its own services.
```
tekul, line 55: I don't think the length used for Google's services matters here, since this is server side code which will be for some custom service.

Crypto/OTP.hs
```diff
@@ -26,6 +26,43 @@
...
+-- > -- RFC 4226 recommends a 160 bit (20 byte) secret key https://tools.ietf.org/html/rfc4226#section-4
+-- > randomTOTPSecretKey :: MonadIO m => m (TOTPSecretKey, GoogleAuthenticatorEncodedKey)
+-- > randomTOTPSecretKey = do
+-- >   key <- liftIO $ fst . randomBytesGenerate 20 <$> getSystemDRG
```
tekul, line 59: This should probably use `MonadRandom` and `getRandomBytes`.

Crypto/OTP.hs
```diff
@@ -26,6 +26,43 @@
...
+-- >   return (TOTPSecretKey key, googleAuthenticatorEncode key)
+-- >
+-- > googleAuthenticatorEncode :: ByteString -> GoogleAuthenticatorEncodedKey
+-- > googleAuthenticatorEncode bs = GoogleAuthenticatorEncodedKey $ stripPadding $ TE.decodeUtf8 $ convertToBase Base32 bs
+-- >   where
+-- >     stripPadding t = T.dropWhileEnd (== '=') t
```
tekul, line 66: My main concern with the code is that it is quite long for what it is showing (essentially just generating and encoding some bytes, which isn't specific to OTP). You could perhaps summarise by saying something like:

> Generate a random 20 byte key using [`getRandomBytes 20`](http://hackage.haskell.org/package/cryptonite-0.25/docs/Crypto-Random.html#v:getRandomBytes), store the key in the database and return the base-32 encoded key to the user for entry into the Google Authenticator application. The key can be encoded using [`convertToBase Base32`](http://hackage.haskell.org/package/memory-0.14.16/docs/Data-ByteArray-Encoding.html#v:convertToBase) from the `memory` package.

I don't really mind about the size of docs in the code, but I think an extended code sample with imports should be compilable (which I don't think this is), and show how to use the functionality of the module in question. For example, sample code to generate a sequence of bytes should probably be in the `Crypto.Random` module and could then be linked from elsewhere.

Crypto/OTP.hs
```diff
@@ -26,6 +26,43 @@
...
+-- > import qualified Data.Text.Encoding as TE
+-- > import Data.ByteArray (ByteArrayAccess) -- memory package
+-- > import Data.ByteArray.Encoding (convertToBase, Base(Base32))
+-- > import Crypto.Random (getSystemDRG, randomBytesGenerate)
+-- >
+-- > newtype TOTPSecretKey = TOTPSecretKey ByteString
+-- >     deriving (Show, ByteArrayAccess)
+-- > 
+-- > newtype GoogleAuthenticatorEncodedKey = GoogleAuthenticatorEncodedKey Text
+-- >   deriving (Show)
```
tekul, line 53: I'd probably avoid deriving `Show` for the secret key types.

tekul (commented):
Thanks for the PR! More docs are always a good thing. I put down a few thoughts which may or may not make sense :-).

vincenthz:
archiving repository

PR #223, opened 2018-03-12
Drop GHC 7.8 and GHC 7.10 support, refer to pkg-guidelines

vincenthz:
8.4 is out, drop long in the teeth compilers


vincenthz:
I left long enough, so at this rate silence = everyone agrees :)

tekul:
The bleeding edge is where it's at, man.

Issue #225, opened 2018-04-03
Extending EllipticCurve type classes


ocheron:
Current API in `Crypto.ECC` is missing at least some features:
- encoding/decoding of scalars
- arithmetic primitives for scalars

#### Serialization of scalars

For NIST curves it is clearly expected as big-endian, however I'm not sure this will necessarily apply to the twisted Edwards curves. Implementation of Ed25519/Ed448 is based on little-endian serialization and this is what libraries provide. For the API, selectable endianness could be considered.

#### Scalar arithmetic

This could be based on `Num` instances for the basic operations.
The only issue is what to do about `abs` and `signum`: either not implement them, or use some criterion like parity as sign.

PR #226, opened 2018-04-03
ECDSA with a type class

ocheron:
Goal here is to get faster ECDSA implementation for P-256:
- removing some side-channels in primitive operations and adding scalar inversion with Fermat's little theorem
- implementing a clone of existing ECDSA API but based on Crypto.ECC

For now the type-class internals are not exposed, waiting for some extension of class hierarchy as explained in #225.


ocheron:
To be split into multiple parts.

Issue #109, opened 2016-11-03
internal / underlying blake2 symbols aren't mangled or private


cartazio:
I'm helping some folks with some crypto heavy code and this change would make our lives much easier.  If I helped get this sorted, would it be possible to turn around a release with that hange pretty fast ?

cartazio:
as a bit of context https://github.com/ocharles/argon2/issues/10 and 

```

duplicate symbol _blake2b_init_key in:
    /Users/carter/.cabal/store/ghc-8.0.1/cryptonite-0.20-0af5802ff1a1deb4cb611283f80d9b336faa42a9e6e78f7dbab10d01ae0d2eeb/lib/libHScryptonite-0.20-0af5802ff1a1deb4cb611283f80d9b336faa42a9e6e78f7dbab10d01ae0d2eeb.a(blake2b.o)
    /Users/carter/.cabal/store/ghc-8.0.1/argon2-1.2.0-b9f6a4cdf7fc9c099739f9e37ad85d14069af01351740db465356c55afa097c1/lib/libHSargon2-1.2.0-b9f6a4cdf7fc9c099739f9e37ad85d14069af01351740db465356c55afa097c1.a(blake2b.o)
duplicate symbol _blake2b_init in:
    /Users/carter/.cabal/store/ghc-8.0.1/cryptonite-0.20-0af5802ff1a1deb4cb611283f80d9b336faa42a9e6e78f7dbab10d01ae0d2eeb/lib/libHScryptonite-0.20-0af5802ff1a1deb4cb611283f80d9b336faa42a9e6e78f7dbab10d01ae0d2eeb.a(blake2b.o)
    /Users/carter/.cabal/store/ghc-8.0.1/argon2-1.2.0-b9f6a4cdf7fc9c099739f9e37ad85d14069af01351740db465356c55afa097c1/lib/libHSargon2-1.2.0-b9f6a4cdf7fc9c099739f9e37ad85d14069af01351740db465356c55afa097c1.a(blake2b.o)
duplicate symbol _blake2b_init_param in:
    /Users/carter/.cabal/store/ghc-8.0.1/cryptonite-0.20-0af5802ff1a1deb4cb611283f80d9b336faa42a9e6e78f7dbab10d01ae0d2eeb/lib/libHScryptonite-0.20-0af5802ff1a1deb4cb611283f80d9b336faa42a9e6e78f7dbab10d01ae0d2eeb.a(blake2b.o)
[...]

PR #227, opened 2018-04-05
Support building against system argon2.

QuLogic:
This echos the option that exists in the `argon2` package to not use the internal copy and build against a system version.

In a way, this is an alternate way to "fix" #109 because there wouldn't be any internal copy of the symbols that could conflict.

### DIFF
--- a/Crypto/KDF/Argon2.hs
+++ b/Crypto/KDF/Argon2.hs
@@ -147,7 +147,7 @@ cOfVariant Argon2id = 2
 csizeOfInt :: Int -> CSize
 csizeOfInt = fromIntegral
 
-foreign import ccall unsafe "cryptonite_argon2_hash"
+foreign import ccall unsafe "cryptonite_argon2_hash_haskell"
     argon2_hash :: Word32 -> Word32 -> Word32
                 -> Ptr Pass -> CSize
                 -> Ptr Salt -> CSize
--- a/cbits/cryptonite_argon2.c
+++ b/cbits/cryptonite_argon2.c
@@ -8,8 +8,14 @@ int cryptonite_argon2_hash_haskell(const uint32_t t_cost,
                                    const size_t hashlen, argon2_type type,
                                    const uint32_t version)
 {
+#ifdef USE_SYSTEM_LIBRARY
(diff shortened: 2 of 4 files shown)


ocheron:
Tested on latest Ubuntu but it does not work:
```
$ cabal configure -fuse-system-library
Resolving dependencies...
Configuring cryptonite-0.25...
cabal: parsing output of pkg-config --modversion failed
```

This is related to haskell/Cabal#163:
```
$ pkg-config --modversion libargon2
0~20161029
```

QuLogic:
That is actually a bug in the Ubuntu/Debian packages; they should not include the `0~` prefix. Can you report a bug to one of them? I do not use either.

QuLogic:
Actually, looks like there's a bug for [Ubuntu](https://bugs.launchpad.net/ubuntu/+source/argon2/+bug/1761418) and [Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=894885) already, opened by the separate argon2 Haskell package developer.

ocheron:
Thanks for pointing to the fixed packages, I tried again with that.

But regarding argon2, I don't see how that could work: the local implementation uses name `cryptonite_argon2_hash` instead of `argon2_hash` and has been simplified.
In commit 253bf0cb8bcc42bb86bb6d68af39d94ef1af371f, parameters `encoded` and `encodedlen` have been removed.
This is not the same API.

Regarding BLAKE2, when I test this gives segmentation faults. The Haskell layer in Crypto.Hash.Blake2* is tied to specific buffer sizes for hashing contexts and this approach will not be compatible with a variable implementation coming from a shared library.

QuLogic:
Hmm, you're right; I think I may have written this before some of those changes. I will get back to you on this.

QuLogic:
For the moment, I've cut this down to just system argon2.

With blake2, it's a bit complicated because they started with [a program](https://github.com/BLAKE2/BLAKE2/), that was sort-of-embeddable, and later wrote [a library](https://github.com/BLAKE2/libb2/), but it has a different API. Cryptonite uses the files from the program, so it would need some sort of compat to work with the library. Changing to embedding the library is also possible, but it re-arranged the configuration options around and I'm not super-sure how to do that here.

vincenthz:
archiving repository

PR #232, opened 2018-04-16
Add Show instance for HMAC type

rtentser:


ocheron:
A `Read` instance should be easy too.

vincenthz:
The reason it doesn't have a `Show` instance, is that it make it less easy to print by mistake. you have to actively want to print it (through a wrapper type or an orphan instance)

PR #238, opened 2018-05-08
Extend the internal interface of the Blowfish module.

lpeterse:
Hi,

I'm currently implementing bcrypt_pbkdf (as used by OpenSSH to encrypt private key files)
and found myself in need of certain low-level operations that are currently not exposed
by the Blowfish module.
After reading the code thoroughly I figured out how it has already been extended
in order to work for the regular BCrypt algorithm, but the eksBlowfish function is not general enough
and I wanted to avoid to just add another couple of extensions to the module instead
of trying to properly generalizing it and splitting separations of concern.
The new module interface is rather similar to that of other implementations, especially the Golang one.
No public interfaces or test cases had to be touched for this change.

I'm just seeking for a little input and feedback on this. I don't expect it to merged right away,
but just want to know whether such a change is generally welcome or rather too drastical.

P.S: This commit improves the performance of the BCrypt calculation by about 11% on my machine.

From the commit message:

In preparation of an implementation of the bcrypt_pbkdf (a
variant of PBKDF2 used by OpenSSH) algorithm,
certain low-level operations of the Blowfish algorithm need to
be generalized and exposed.

The Blowfish.Primitive module has already been extended to
account for the requirements imposed by the BCrypt algorithm,
but the salt length was limited to 16 bytes and the BCrypt
specific key schedule setup has been hard-coded into the Blowfish
module.

This commit makes a clear distintion between the expandKey and
expandKeyWithSalt operation. Both take arbitrary sized salts
and keys now. The specialized operation for 16 byte salts as used
by BCrypt has been preserved and is selected automatically.
Also, the BCrypt specific parts have been move to the BCrypt
module with regard to separation of concern.

A benchmark for generating BCrypt hashes with cost 10 shows a
performance improvement from 158 to 141ms on average (Intel i5-6500)
after this refactoring.
Further experiments suggest that the specialized expandKeyWithSalt128
does not have any advantage over the generalized version
and might be removed in favour of less branches and exceptional
behaviour.

### DIFF
--- a/Crypto/Cipher/Blowfish/Box.hs
+++ b/Crypto/Cipher/Blowfish/Box.hs
@@ -5,15 +5,19 @@
 -- Portability : Good
 {-# LANGUAGE MagicHash #-}
 module Crypto.Cipher.Blowfish.Box
-    ( createKeySchedule
+    (   KeySchedule(..)
+    ,   createKeySchedule
     ) where
 
-import Crypto.Internal.WordArray (mutableArray32FromAddrBE, MutableArray32)
+import           Crypto.Internal.WordArray (MutableArray32,
+                                            mutableArray32FromAddrBE)
+
+newtype KeySchedule = KeySchedule MutableArray32
 
 -- | Create a key schedule mutable array of the pbox followed by
 -- all the sboxes.
-createKeySchedule :: IO MutableArray32
-createKeySchedule = mutableArray32FromAddrBE 1042 "\
+createKeySchedule :: IO KeySchedule
+createKeySchedule = KeySchedule `fmap` mutableArray32FromAddrBE 1042 "\
     \\x24\x3f\x6a\x88\x85\xa3\x08\xd3\x13\x19\x8a\x2e\x03\x70\x73\x44\
     \\xa4\x09\x38\x22\x29\x9f\x31\xd0\x08\x2e\xfa\x98\xec\x4e\x6c\x89\
     \\x45\x28\x21\xe6\x38\xd0\x13\x77\xbe\x54\x66\xcf\x34\xe9\x0c\x6c\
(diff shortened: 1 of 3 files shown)


ocheron:
The transformed implementation looks good.

For BCrypt with GHC 8.6.3, performance is further improved when adding a bang pattern to the first argument `i` of `doRound` inner functions. By something like 30%, not only 11%. And for Blowfish128 the PR gives a good performance improvement too.

Before:
```
block-cipher/ECB/Blowfish128-input=1024  mean 51.97 μs  ( +- 764.5 ns  )
bcrypt/cryptonite-BCrypt-4               mean 2.566 ms  ( +- 21.72 μs  )
bcrypt/cryptonite-BCrypt-5               mean 4.949 ms  ( +- 27.13 μs  )
bcrypt/cryptonite-BCrypt-7               mean 19.21 ms  ( +- 76.61 μs  )
bcrypt/cryptonite-BCrypt-11              mean 304.3 ms  ( +- 595.1 μs  )
```

After (PR):
```
block-cipher/ECB/Blowfish128-input=1024  mean 16.48 μs  ( +- 161.8 ns  )
bcrypt/cryptonite-BCrypt-4               mean 2.265 ms  ( +- 14.05 μs  )
bcrypt/cryptonite-BCrypt-5               mean 4.430 ms  ( +- 50.78 μs  )
bcrypt/cryptonite-BCrypt-7               mean 17.35 ms  ( +- 103.6 μs  )
bcrypt/cryptonite-BCrypt-11              mean 275.3 ms  ( +- 338.0 μs  )
```

After (PR modified with bang pattern):
```
block-cipher/ECB/Blowfish128-input=1024  mean 16.46 μs  ( +- 102.1 ns  )
bcrypt/cryptonite-BCrypt-4               mean 1.711 ms  ( +- 9.857 μs  )
bcrypt/cryptonite-BCrypt-5               mean 3.324 ms  ( +- 19.37 μs  )
bcrypt/cryptonite-BCrypt-7               mean 13.01 ms  ( +- 75.34 μs  )
bcrypt/cryptonite-BCrypt-11              mean 206.2 ms  ( +- 308.7 μs  )
```

PR #240, opened 2018-05-10
Add bcrypt_pbkdf key derivation function

lpeterse:
This is an implementation of the [bcrypt_pbkdf function from OpenBSD](http://man.openbsd.org/bcrypt_pbkdf.3).

It is a slight variation of the normal bcrypt algorithm. Among other uses, it is used to derive the encryption key for OpenSSL private key files.

This PR is based on my previous one that prepared the Blowfish implementation.

### DIFF
--- a/Crypto/KDF/BCryptPBKDF.hs
+++ b/Crypto/KDF/BCryptPBKDF.hs
@@ -1,0 +1,187 @@
+-- |
+-- Module      : Crypto.KDF.BCryptPBKDF
+-- License     : BSD-style
+-- Stability   : experimental
+-- Portability : Good
+--
+-- Port of the bcrypt_pbkdf key derivation function from OpenBSD
+-- as described at <http://man.openbsd.org/bcrypt_pbkdf.3>.
+module Crypto.KDF.BCryptPBKDF
+    ( Parameters (..)
+    , generate
+    , hashInternal
+    )
+where
+
+import           Basement.Block                   (MutableBlock)
+import qualified Basement.Block                   as Block
+import qualified Basement.Block.Mutable           as Block
+import           Basement.Monad                   (PrimState)
+import           Basement.Types.OffsetSize        (CountOf (..), Offset (..))
+import           Control.Exception                (finally)
+import           Control.Monad                    (when)
+import qualified Crypto.Cipher.Blowfish.Box       as Blowfish
+import qualified Crypto.Cipher.Blowfish.Primitive as Blowfish
+import           Crypto.Hash.Algorithms           (SHA512 (..))
+import           Crypto.Hash.Types                (Context,
+                                                   hashDigestSize,
+                                                   hashInternalContextSize,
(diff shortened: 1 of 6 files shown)


Crypto/KDF/BCryptPBKDF.hs
```diff
@@ -0,0 +1,188 @@
...
+                        memXor outPtr outPtr tmpPtr outLen
+                    -- Spread the current out buffer evenly over the key buffer.
+                    -- After both loops have run every byte of the key buffer
+                    -- will have been written to exactly once, but not all bytes
+                    -- of the output will have been used.
+                    forM_ [0..fromIntegral outLen - 1] $ \outIdx-> do
+                        let keyIdx = outIdx * blocks + block - 1
+                        when (outIdx < outLen) $ do
+                            w8 <- peekByteOff outPtr outIdx :: IO Word8
+                            pokeByteOff keyPtr keyIdx w8
```
ocheron, line 134: New condition `outIdx < outLen` looks incorrect, I think it should be `keyIdx < keyLen`.
Also `fromIntegral` is not needed for `outLen`.

Crypto/KDF/BCryptPBKDF.hs
```diff
@@ -0,0 +1,188 @@
...
+hashInternalMutable :: (B.ByteArrayAccess pass, B.ByteArrayAccess salt)
+    => Blowfish.KeySchedule
+    -> pass
+    -> salt
+    -> MutableBlock Word8 (PrimState IO)
+    -> IO ()
+hashInternalMutable bfks passHash saltHash outMBlock
+    | B.length passHash                  /= 64 = error "passHash must be 512 bits"
+    | B.length saltHash                  /= 64 = error "saltHash must be 512 bits"
+    | Block.mutableLengthBytes outMBlock /= 32 = error "outMBlock must be 256 bits"
```
ocheron: Testing `outMBlock` length is not very useful, it's always allocated inside this module.
Verifying `passHash` and `saltHash` could be moved to `hashInternal`.

ocheron (commented):
Nice to see something with basement API for a change.
Can you rebase on master, look at the two points I spotted, and remove unneeded imports?

lpeterse:
Thanks for merging my other PR.
I'll rebase and fix this one in the next few days..

Crypto/KDF/BCryptPBKDF.hs
lpeterse: Correct.

Crypto/KDF/BCryptPBKDF.hs
lpeterse: Also correct.

lpeterse:
I fixed the issues you found and combined everything in a single commit as history got messed up anyway. It is rebased on current master.

ocheron:
Looks good, thank you.

PR #241, opened 2018-06-05
add ECDSA sign/verify digest APIs

wangbj:
Is it possible to add few APIs to allow sign/verify digest instead of message? This allows sign/verify digests created by different party (not necessarily cryptonite), without introducing any complexity.


ocheron:
Looking good but `signWithDigest` should be changed to `signDigestWith`.

Also I think we can drop the `hash` argument where `Digest hash` is available.
Maybe except for `tHashDigest`, which can be called internally with `undefined`.

ocheron:
Rebased and merged as part of #270 with additional changes.

PR #242, opened 2018-06-18
Derive a Data instance for Digest.

LeifW:


LeifW:
Since this is a newtype wrapper around `Block`, in my own code for now I just copied the impl for the `Data` instance for `Block`:
```haskell
instance Typeable a => Data (Digest a) where
    gunfold k z c = error "gunfold"
    toConstr _ = error "toConstr"
    dataTypeOf _ = mkNoRepType "Crypto.Hash.Types.Digest"
```

vincenthz:
Thanks !

PR #247, opened 2018-09-06
Add Rabin (and variants) cryptosystem

crodriguezvega:
I have implemented the basic Rabin cryptosystem, and a couple of its variants (including Rabin-Williams). There are, however, a couple of topics for which I need some direction:

- The encryption algorithms do not use any padding scheme at the moment. I guess it would make sense to use something like OAEP, but at the moment it looks like it is coupled with the RSA implementation. Would it make sense to refactor the current implementation of OAEP to make it independent of the encryption algorithm? Or should I just implement my own version of OAEP (or another padding scheme for that matter) for Rabin?
- I have not been able to found 3rd-party test vectors for Rabin, so at the moment I have just added a few tests that test the whole generate key -> encrypt/sign -> decrypt/verify cycle. Is this acceptable or should I create my own test vectors instead?

Looking forward to receiving some feedback.


vincenthz:
a bit short on time, excuse the conciseness:

* I'm not sure OAEP can be easily refactored from RSA, so happy with duplicating the bit needed.
* If you really can't find any test vectors, then at least having some hardcoded one from this implementation, make it less likely it changes by mistake, not ideal but better than nothing.

crodriguezvega:
@vincenthz Thank you for the answers. I will work on these two points.

crodriguezvega:
I pretty much duplicated the OAEP implementation so that I can use it in the Rabin cryptosystems and I also created some test vectors to have some basic checks on the implementations.

cryptonite.cabal
```diff
@@ -231,6 +236,7 @@ Library
 
   Build-depends:     bytestring
                    , memory >= 0.14.14
+                   , random
```
vincenthz: I couldn't find any reference to `random`, mistake ?

crodriguezvega: I looked in the history of the file on master and it seems like that reference has not been there (at least) since I created my branch... If you think it is still needed, then I am fine with adding it, of course.

vincenthz: Sorry if I wasnt clear, it just appear random is there as a new dependency, but doesnt seems used anywhere. Generally I dont want to have to depend on new things and random is no exception (specially its a particular bad random source and generator)

crodriguezvega: Sorry, I interpreted your comment the wrong way. You're right: the reference to `random` is part of a leftover that I should have removed. It is fixed now.

vincenthz: thanks !

vincenthz:
a bit slow to integrate this (on my side), but we made it. thanks !

crodriguezvega:
@vincenthz Thanks a lot for the review!

PR #248, opened 2018-09-11
Add SHA512Half hash algorithm

gip:
This PR adds the `SHA512Half` hash algorithm that is used in the crypto space. It also export `HashAlgorithm(..)` to make it possible to add new hash algorithms that works with cryptonite.


ocheron:
FYI the class is already exported from module `Crypto.Hash.IO` to implement custom algorithms.

vincenthz:
if you need to define a new hash algorithm, there's a `gen` directory for all the hashes.

I'm a bit dubious about the usefulness of this, considering it's just SHA512 with the output truncated compared to SHA512t-256 which is a proper cryptographic construction (and which have an implication on the initial value).

Would it be not better to export a `Digest SHA512 -> Digest SHA512Half` or maybe a more generic construction:

```data TruncatedDigest (n :: Nat) hash = ...```

so that you can have `TruncatedDigest 256 SHA512` ?

vincenthz:
going to closing this.. since I don't think it's particularly needed and no further engagement from @gip

PR #262, opened 2019-01-13
RSA-PSS with arbitrary key size

ocheron:
As reported in vincenthz/hs-tls#333, implementation of RSA-PSS does not handle a key with length not multiple of 8 bits. Available data type does not store the bit length, but like @vdukhovni demonstrated we can use `numBits` to get it. The PR adds this for both generation and verification, and adds test vectors with unusual key size.


Crypto/PubKey/RSA/PSS.hs
```diff
@@ -161,30 +163,37 @@ verifyDigest :: HashAlgorithm hash
              -> ByteString  -- ^ Signature
              -> Bool
 verifyDigest params pk digest s
-    | public_size pk /= B.length s        = False
+    | B.length s /= k                     = False
+    | B.any (/= 0) pre                    = False
```
vdukhovni, line 167: This form `B.any (/= 0) bytes` is IMHO cleaner than the below `not (B.all (== 0) bytes)` used in the `ps0` case.  Perhaps change that one to match?

Crypto/PubKey/RSA/PSS.hs
```diff
@@ -161,30 +163,37 @@ verifyDigest :: HashAlgorithm hash
...
     | b1 /= B.singleton 1                 = False
-    | otherwise                           = h == B.convert h'
+    | otherwise                           = B.eq h h'
         where -- parameters
               hashLen   = hashDigestSize (pssHash params)
               mHash     = B.convert digest
-              dbLen     = public_size pk - hashLen - 1
-              pubBits   = public_size pk * 8 -- to change if public_size is converted in bytes
+              k         = public_size pk
+              emLen     = if emTruncate pubBits then k - 1 else k
```
vdukhovni, line 176: Why not just ```emLen = (pubBits - 1) `div` 8```? If we want a consistency sanity check between `pubBits` and `k` it could  be ```k == (pubBits + 7) `div` 8```, if there's ever any doubt that the `public_size` is consistent with the modulus.  Note that IIRC, unfortunately, the library does not hide the RSA key structure internals, and provides no constructors that ensure correctness of the provided values.  The application gets to set the modulus and `public_size` as it pleases.  This appears to be difficult to fix without a breaking change to the library, but should perhaps be done at some point, bumping the major number, and fixing all the dependencies in the X.509 and TLS libraries.

Crypto/PubKey/RSA/PSS.hs
```diff
@@ -69,18 +70,19 @@ signDigestWithSalt :: HashAlgorithm hash
                    -> Digest hash   -- ^ Message digest
                    -> Either Error ByteString
 signDigestWithSalt salt blinder params pk digest
-    | k < hashLen + saltLen + 2 = Left InvalidParameters
-    | otherwise                 = Right $ dp blinder pk em
+    | emLen < hashLen + saltLen + 2 = Left InvalidParameters
+    | otherwise                     = Right $ dp blinder pk em
     where k        = private_size pk
+          emLen    = if emTruncate pubBits then k - 1 else k
```
vdukhovni, line 76: See comments re`emTruncate` below.

Crypto/PubKey/RSA/PSS.hs
```diff
@@ -161,30 +163,37 @@ verifyDigest :: HashAlgorithm hash
...
               db        = B.pack $ normalizeToKeySize pubBits $ B.zipWith xor maskedDB dbmask
               (ps0,z)   = B.break (== 1) db
               (b1,salt) = B.splitAt 1 z
               m'        = B.concat [B.replicate 8 0,mHash,salt]
               h'        = hashWith (pssHash params) m'
 
+-- When the modulus has bit length 1 modulo 8 we drop the first byte.
+emTruncate :: Int -> Bool
+emTruncate bits = ((bits-1) .&. 0x7) == 0
+
```
vdukhovni, line 193: I would not have implemented this function, per the comment above.

tests/KAT_PubKey/PSS.hs
```diff
@@ -470,4 +339,10 @@ pssTests = testGroup "RSA-PSS"
         [ doVerifyTest rsaKeyInt (katZero, vectorInt) ]
     , testGroup "signature key 1024" $ map (doSignTest rsaKey1) (zip [katZero..] vectorsKey1)
     , testGroup "verify key 1024" $ map (doVerifyTest rsaKey1) (zip [katZero..] vectorsKey1)
+    , testGroup "signature key 1025" $ map (doSignTest rsaKey2) (zip [katZero..] vectorsKey2)
+    , testGroup "verify key 1025" $ map (doVerifyTest rsaKey2) (zip [katZero..] vectorsKey2)
+    , testGroup "signature key 1026" $ map (doSignTest rsaKey3) (zip [katZero..] vectorsKey3)
+    , testGroup "verify key 1026" $ map (doVerifyTest rsaKey3) (zip [katZero..] vectorsKey3)
+    , testGroup "signature key 1031" $ map (doSignTest rsaKey8) (zip [katZero..] vectorsKey8)
+    , testGroup "verify key 1031" $ map (doVerifyTest rsaKey8) (zip [katZero..] vectorsKey8)
```
vdukhovni, line 347: While we're at it, perhaps just run through all the test vectors from RSA for the 10 key sizes?
```
# Key  1: 1024 bits
# Key  2: 1025 bits
# Key  3: 1026 bits
# Key  4: 1027 bits
# Key  5: 1028 bits
# Key  6: 1029 bits
# Key  7: 1030 bits
# Key  8: 1031 bits
# Key  9: 1536 bits
# Key 10: 2048 bits
```

vdukhovni (commented):
Many thanks for this. Looks correct to me, just some nits that you might care to fix or dismiss.

Crypto/PubKey/RSA/PSS.hs
ocheron: Done with 1d5947f.

tests/KAT_PubKey/PSS.hs
ocheron: Initially I did all examples. But module was growing large, so I removed some vectors not really necessary.

ocheron:
Thanks for the review.

I agree there's a bit of a concern if `public_size` is not consistent with the bit length of the modulus.
For now I preferred to keep using `public_size` as much as possible.
It is the length returned by the primitives `ep` and `dp`.

Issue #264, opened 2019-01-26
DSA does not truncate hashes correctly


crockeea:
`sign` does not truncate long hash outputs as specified in FIPS 186-4, but `verify` *does* truncate the hash. This means that if the hash digest size is larger than `q`, verification of signatures will fail (even when `q` is a multiple of 8).

crockeea:
Fixed by PR #265.

PR #265, opened 2019-01-26
Fixed hash truncation bug in DSA

crockeea:
This fixes the hash truncation bug from #264. I added KATs from RFC 6979 to verify correctness.

DSA should also now handle `q` whose bit length is not a multiple of 8. I was unable to find any KATs for parameters of this type, so this code is untested.


ocheron:
Code is correct and the test vectors are definitely wanted.

One concern though: what you add is really similar to function `tHash` in Crypto.PubKey.ECC.ECDSA.
So one may also move `tHash` somewhere like Crypto.PubKey.Internal, and use this for both DSA and ECDSA. Would you like trying this?

Also unclear to me why you kept the FIXME.

crockeea:
@ocheron Thanks for the review. I'll fix your suggestions soon.

crockeea:
Should be fixed now.

ocheron (approved):
Yes, looking good. Thanks.

PR #266, opened 2019-01-30
Copyedit docs

chris-martin:
I went through the whole package attempting to clean up the docs a bit. Most of the changes are for consistency in capitalization and periods. There are also a handful of fixed misspellings, grammatical errors, and places where Haddock wasn't formatting things quite right.

I hope this is welcome :smiley: Sorry it's a big commit; if there are conflicts with anybody else's active work I'll be happy to split this up to avoid stepping on anybody's toes.

### DIFF
--- a/Crypto/Cipher/AES/Primitive.hs
+++ b/Crypto/Cipher/AES/Primitive.hs
@@ -194,31 +194,31 @@ initAES k
         aesInit ptr = withByteArray k $ \ikey ->
             c_aes_init (castPtr ptr) (castPtr ikey) (fromIntegral len)
 
--- | encrypt using Electronic Code Book (ECB)
+-- | Encrypt using Electronic Code Book (ECB)
 {-# NOINLINE encryptECB #-}
 encryptECB :: ByteArray ba => AES -> ba -> ba
 encryptECB = doECB c_aes_encrypt_ecb
 
--- | encrypt using Cipher Block Chaining (CBC)
+-- | Encrypt using Cipher Block Chaining (CBC)
 {-# NOINLINE encryptCBC #-}
 encryptCBC :: ByteArray ba
            => AES        -- ^ AES Context
            -> IV AES     -- ^ Initial vector of AES block size
-           -> ba         -- ^ plaintext
-           -> ba         -- ^ ciphertext
+           -> ba         -- ^ Plaintext
+           -> ba         -- ^ Ciphertext
 encryptCBC = doCBC c_aes_encrypt_cbc
 
--- | generate a counter mode pad. this is generally xor-ed to an input
+-- | Generate a counter mode pad. This is generally xor-ed to an input
 -- to make the standard counter mode block operations.
 --
--- if the length requested is not a multiple of the block cipher size,
+-- If the length requested is not a multiple of the block cipher size,
 -- more data will be returned, so that the returned bytearray is
 -- a multiple of the block cipher size.
 {-# NOINLINE genCTR #-}
 genCTR :: ByteArray ba
-       => AES    -- ^ Cipher Key.
-       -> IV AES -- ^ usually a 128 bit integer.
-       -> Int    -- ^ length of bytes required.
+       => AES    -- ^ Cipher key
+       -> IV AES -- ^ Usually a 128 bit integer
+       -> Int    -- ^ Length of bytes required
        -> ba
 genCTR ctx (IV iv) len
     | len <= 0  = B.empty
@@ -227,14 +227,14 @@ genCTR ctx (IV iv) len
         (nbBlocks',r) = len `quotRem` 16
         nbBlocks = if r == 0 then nbBlocks' else nbBlocks' + 1
 
--- | generate a counter mode pad. this is generally xor-ed to an input
+-- | Generate a counter mode pad. This is generally xor-ed to an input
 -- to make the standard counter mode block operations.
 --
--- if the length requested is not a multiple of the block cipher size,
+-- If the length requested is not a multiple of the block cipher size,
 -- more data will be returned, so that the returned bytearray is
 -- a multiple of the block cipher size.
 --
--- Similiar to 'genCTR' but also return the next IV for continuation
+-- Similiar to 'genCTR' but also returns the next IV for continuation.
 {-# NOINLINE genCounter #-}
 genCounter :: ByteArray ba
            => AES
@@ -257,15 +257,15 @@ genCounter ctx iv len
  - RULES "snd . genCounter" forall ctx iv len .  snd (genCounter ctx iv len) = genCTR ctx iv len
  -}
 
--- | encrypt using Counter mode (CTR)
+-- | Encrypt using Counter mode (CTR)
 --
--- in CTR mode encryption and decryption is the same operation.
+-- In CTR mode encryption and decryption is the same operation.
 {-# NOINLINE encryptCTR #-}
 encryptCTR :: ByteArray ba
            => AES        -- ^ AES Context
-           -> IV AES     -- ^ initial vector of AES block size (usually representing a 128 bit integer)
-           -> ba         -- ^ plaintext input
-           -> ba         -- ^ ciphertext output
+           -> IV AES     -- ^ Initial vector of AES block size (usually representing a 128 bit integer)
+           -> ba         -- ^ Plaintext input
+           -> ba         -- ^ Ciphertext output
 encryptCTR ctx iv input
     | len <= 0          = B.empty
     | B.length iv /= 16 = error $ "AES error: IV length must be block size (16). Its length is: " ++ (show $ B.length iv)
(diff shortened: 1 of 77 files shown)


tekul:
Sounds good. I'll try and have a look through sometime next week.

vincenthz:
Just was looking at this. Not sure this is good as a mega commit (I supposed the merge conflict now kind of goes in this direction).

Just overlooking, but not sure about "4 bytes constant" -> "4 byte constant".

tekul:
Oops. I clean forgot about this, sorry. A "4 byte constant" makes sense. Like a "32 bit word".

chris-martin:
Resolved conflicts.

Crypto/Cipher/AES/Primitive.hs
```diff
@@ -432,20 +432,20 @@ ocbInit ctx iv = unsafeDoIO $ do
...
             withByteArray input $ \i ->
             c_aes_ocb_aad ocbStPtr aesPtr i (fromIntegral $ B.length input)
 
--- | append data to encrypt and append to the OCB context
+-- | Append data to encrypt and append to the OCB context.
 --
--- the bytearray needs to be a multiple of the AES block size, unless it's the last call to this function.
--- need to happen after AAD appending, or after initialization if no AAD data.
+-- The bytearray needs to be a multiple of the AES block size, unless it's the last call to this function.
+-- Needs to happen after AAD appending, or after initialization if no AAD data.
```
tekul, line 448: I think this would be better as "Needs to happen after appending the AAD, or after initialization if there is no AAD data."

Crypto/Cipher/AES/Primitive.hs
```diff
@@ -455,10 +455,10 @@ ocbAppendEncrypt ctx ocb input = unsafeDoIO $ withOCBKeyAndCopySt ctx ocb doEnc
             withByteArray input $ \i ->
             c_aes_ocb_encrypt (castPtr o) ocbStPtr aesPtr i (fromIntegral len)
 
--- | append data to decrypt and append to the OCB context
+-- | Append data to decrypt and append to the OCB context.
 --
--- the bytearray needs to be a multiple of the AES block size, unless it's the last call to this function.
--- need to happen after AAD appending, or after initialization if no AAD data.
+-- The bytearray needs to be a multiple of the AES block size, unless it's the last call to this function.
+-- Need to happen after AAD appending, or after initialization if no AAD data.
```
tekul, line 461: `s/Need/Needs/`. Also the previous comment above (since this is the same text).

Crypto/Cipher/AES/Primitive.hs
```diff
@@ -505,19 +505,19 @@ ccmInit ctx iv n m l
...
 ccmAppendAAD ctx ccm input = unsafeDoIO $ snd <$> withCCMKeyAndCopySt ctx ccm doAppend
   where doAppend ccmStPtr aesPtr =
             withByteArray input $ \i -> c_aes_ccm_aad ccmStPtr aesPtr i (fromIntegral $ B.length input)
 
--- | append data to encrypt and append to the CCM context
+-- | Append data to encrypt and append to the CCM context
 --
--- the bytearray needs to be a multiple of AES block size, unless it's the last call to this function.
--- needs to happen after AAD appending, or after initialization if no AAD data.
+-- The bytearray needs to be a multiple of AES block size, unless it's the last call to this function.
```
tekul, line 519: `s/AES block size/the AES block size/`

Crypto/Cipher/Types/Block.hs
```diff
@@ -72,59 +72,61 @@ class Cipher cipher => BlockCipher cipher where
 
     -- | Encrypt blocks
     --
-    -- the input string need to be multiple of the block size
+    -- The input string needs to be multiple of the block size.
```
tekul, line 75: `s/needs to be multiple/needs to be a multiple/`

Crypto/Cipher/Types/Block.hs
```diff
@@ -72,59 +72,61 @@ class Cipher cipher => BlockCipher cipher where
...
     ecbEncrypt :: ByteArray ba => cipher -> ba -> ba
 
     -- | Decrypt blocks
     --
-    -- the input string need to be multiple of the block size
+    -- The input string needs to be multiple of the block size.
```
tekul, line 80: `s/needs to be multiple/needs to be a multiple/`

Crypto/PubKey/ECC/DH.hs
```diff
@@ -40,7 +40,7 @@ calculatePublic curve d = q
     q = pointMul curve d g
 
 -- | Generating a shared key using our private number and
---   the other party public point.
+-- the other party public point.
```
tekul, line 43: `s/party/party's/` ?

Crypto/PubKey/ECC/P256.hs
```diff
@@ -130,11 +130,11 @@ pointDh scalar p =
             ccryptonite_p256_points_mul_vartime nzero n px py dx dy
         ccryptonite_p256_to_bin (castPtr dx) dst
 
--- | multiply the point @p with @n2 and add a lifted to curve value @n1
+-- | Multiply the point @p with @n2 and add a lifted to curve value @n1
```
tekul, line 133: We should fix the incomplete markup here. `@p@`, `@n2@` etc.

Crypto/PubKey/ElGamal.hs
```diff
@@ -70,39 +70,39 @@ generateEphemeral q = toEphemeral <$> generatePrivate q
...
--- | sign a message with an explicit k number
+-- | Sign a message with an explicit /k/ number
 --
--- if k is not appropriate, then no signature is returned.
+-- If /k/ is not appropriate, then no signature is returned.
 --
--- with some appropriate value of k, the signature generation can fail,
--- and no signature is returned. User of this function need to retry
--- with a different k value.
+-- With some appropriate value of /k/, the signature generation can fail,
```
tekul, line 97: The use of "appropriate" seems weird here. Could we just say "For some values of k" ?

Crypto/PubKey/RSA.hs
```diff
@@ -77,10 +77,10 @@ generateWith (p,q) size e =
                             , private_qinv = qinv
                             }
 
--- | generate a pair of (private, public) key of size in bytes.
+-- | Generate a pair of (private, public) key of size in bytes.
```
tekul, line 80: Since the order is actually the other way round, maybe change this to "Generate a (public, private) key pair of the given size in bytes".

Crypto/PubKey/RSA.hs
```diff
@@ -97,9 +97,9 @@ generate size e = loop
         q <- generatePrime (8 * (size - (size `div` 2)))
         if p == q then generateQ p else return q
 
--- | Generate a blinder to use with decryption and signing operation
+-- | Generate a blinder to use with decryption and signing operation.
 --
--- the unique parameter apart from the random number generator is the
+-- The unique parameter apart from the random number generator is the
```
tekul, line 102: This is out of date since the random number generator is no longer a parameter (it refers to a pre-`MonadRandom` version). What about

"Generate a blinder from the given RSA N parameter for use with decryption and signing operations" ?

Crypto/PubKey/RSA/Prim.hs
```diff
@@ -18,14 +18,14 @@ import           Crypto.Number.ModArithmetic (expFast, expSafe)
 import           Crypto.Number.Serialize (os2ip, i2ospOf_)
 import           Crypto.Internal.ByteArray (ByteArray)
 
-{- dpSlow computes the decrypted message not using any precomputed cache value.
-   only n and d need to valid. -}
+{- Compute the decrypted message not using any precomputed cache value.
```
tekul, line 21: I think these functions are internal only, and they don't have Haddock comments, so  it doesn't seem worth adding markup here.

Crypto/Random.hs
```diff
@@ -85,7 +85,7 @@ drgNewTest = initializeWords
 
 -- | Generate @len random bytes and mapped the bytes to the function @f.
 --
--- This is equivalent to use Control.Arrow 'first' with 'randomBytesGenerate'
+-- This is equivalent to use Control.Arrow 'first' with 'randomBytesGenerate'.
```
tekul, line 88: `s/to use/to using/`

Crypto/Random/Entropy/Unsafe.hs
```diff
@@ -14,13 +14,13 @@ import Data.Word (Word8)
...
 
--- | Refill the entropy in a buffer
+-- | Refill the entropy in a buffer.
 --
 -- Call each entropy backend in turn until the buffer has
 -- been replenished.
 --
 -- If the buffer cannot be refill after 3 loopings, this will raise
--- an User Error exception
+-- an User Error exception.
```
tekul, line 23: `s/an User/a User/`

tekul (commented):
This looks good to me. It's mostly just standardizing the format of comments and fixing some Haddock markup, typos and the odd spot of franglais :slightly_smiling_face: . I added some comments but a lot of them aren't specifically about these changes, but rather about possible additional issues/improvements that I spotted while reading the changes. Could you have a look at my comments @vincenthz ?

Sorry again for forgetting about this @chris-martin :thinking: .

vincenthz:
archiving repository

PR #277, opened 2019-05-19
Little-endian integer serialization

ocheron:
Adds utilities to serialize/deserialize in reverse order.
This is used with some cryptographic schemes.


vincenthz:
looking good. as a side note on the subject, might be a good idea to go away from `i2osp` `os2ip` someday, with something that speak by itself instead of this old naming scheme.

PR #278, opened 2019-05-19
Faster AES GCM with PCLMULQDQ

ocheron:
Complements the AESNI implementation with CLMUL intrinsics when supported by the CPU.

Before:
```
AE/ChaChaPoly1305                        mean 7.682 μs  ( +- 168.9 ns  )
AE/AES-GCM                               mean 14.32 μs  ( +- 66.21 ns  )
AE/AES-CCM                               mean 21.58 μs  ( +- 96.51 ns  )
```

After (support_pclmuldq disabled):
```
AE/ChaChaPoly1305                        mean 7.728 μs  ( +- 199.7 ns  )
AE/AES-GCM                               mean 14.52 μs  ( +- 94.97 ns  )
AE/AES-CCM                               mean 21.70 μs  ( +- 199.7 ns  )
```

After (support_pclmuldq enabled):
```
AE/ChaChaPoly1305                        mean 7.698 μs  ( +- 170.2 ns  )
AE/AES-GCM                               mean 3.037 μs  ( +- 52.41 ns  )
AE/AES-CCM                               mean 21.65 μs  ( +- 151.6 ns  )
```

For now flag `support_pclmuldq` must still be enabled explicitely.

### DIFF
--- a/Crypto/Cipher/Types/Base.hs
+++ b/Crypto/Cipher/Types/Base.hs
@@ -22,6 +22,7 @@ module Crypto.Cipher.Types.Base
 import           Data.Word
 import           Crypto.Internal.ByteArray (Bytes, ByteArrayAccess, ByteArray)
 import qualified Crypto.Internal.ByteArray as B
+import           Crypto.Internal.DeepSeq
 import           Crypto.Error
 
 -- | Different specifier for key size in bytes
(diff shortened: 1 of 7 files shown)


vincenthz:
much needed improvement, lgtm !

ocheron:
Thanks Vincent. Yes the techniques are a bit complex to understand but result is worth the effort :-)

kazu-yamamoto:
When will be the next version with `support_pclmuldq` enabled released?

kazu-yamamoto:
Why does cryptonite-0.27 still disable this flag?

ocheron:
It's best not to change everything in the same release.
But this can be tried later. Enabling AESNI on Windows too.

kazu-yamamoto:
@ocheron BTW, do you know that kazuho (my friend) found a way to improve the performance of AES GCM for short objects? 

Blogs are here. Hoping that the translation works well:

http://blog.kazuhooku.com/2020/06/quicaes-gcm-12.html

Do you think that `cryptonite` can support the approach?

ocheron:
Something like this should be driven by measurements.
Our implementation is very likely to be suboptimal, there can be other factors to improve first.

PR #296, opened 2019-09-16
Num instances for ECC scalars

ocheron:
Adds missing scalar primitives as well as `Num` instances. Except `abs` and `signum` are not implemented because there could be multiple conventions for them.

### DIFF
--- a/Crypto/ECC/Edwards25519.hs
+++ b/Crypto/ECC/Edwards25519.hs
@@ -64,6 +64,7 @@ module Crypto.ECC.Edwards25519
     -- * Arithmetic functions
     , toPoint
     , scalarAdd
+    , scalarSub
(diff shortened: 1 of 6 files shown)


vincenthz:
`Num` instance makes me uneasy due to fitting too loosely the scalar group (sign, abs, ..). Would it make more sense to either:

* use basement arithmetic tower; That solves the issue of loose interface, that comes with its own set of problems
* add a cryptonite typeclass for group operation that we can keep more reasonable at the expense of having to use different operators ?

The rest looks fine.

ocheron:
Thank you. I think I gave a look to basement numerical classes but had same doubt. Because of the extra operations with polymorphic parameters. Opened haskell-foundation/foundation#529.

PR #307, opened 2020-01-18
Add 64-bit implementation for P256

ocheron:
Speed increase using 64-bit limbs:
- few modifications needed for the p256_int code, mainly changing type and macro definitions
- field-element code moved to conditionally-included file, and rewritten the uint64_t version using exactly the same techniques as uint32_t

Before:
```
P256/pointAddTwoMuls-P256                mean 718.7 μs  ( +- 1.707 μs  )
P256/pointAdd-P256                       mean 28.51 μs  ( +- 237.5 ns  )
P256/pointMul-P256                       mean 345.2 μs  ( +- 810.9 ns  )
```

After:
```
P256/pointAddTwoMuls-P256                mean 434.2 μs  ( +- 967.3 ns  )
P256/pointAdd-P256                       mean 15.98 μs  ( +- 26.63 ns  )
P256/pointMul-P256                       mean 208.7 μs  ( +- 440.7 ns  )
```


cbits/include64/p256/p256_gf.h
```diff
@@ -0,0 +1,713 @@
...
+typedef uint64_t u64;
+typedef int64_t s64;
+typedef __uint128_t u128;
+
+/* Our field elements are represented as five 64-bit limbs.
+ *
+ * The value of an felem (field element) is:
+ *   x[0] + (x[1] * 2**51) + (x[2] * 2**103) + ... + (x[4] * 2**206)
+ *
+ * That is, each limb is alternately 51 or 52-bits wide in little-endian
```
vincenthz, line 54: naively I'ld imagine that it would make more sense to target a uniform unsaturated 52 bit only on 64 bits, instead of going to the same 28-29 alternative implementation than 32 bits, but I haven't looked in details enough. non uniform representation really spooks me TBH.

vincenthz (approved):
nice work once again, the resulting benchmark makes for an impressive speedup !

cbits/include64/p256/p256.h
```diff
@@ -0,0 +1,167 @@
...
+// n' such as n * n' = -1 mod (2^64)
+#define P256_MONTGOMERY_FACTOR 0xCCD1C8AAEE00BC4F
+
+#define P256_LITERAL(lo,hi) (((uint32_t) (lo)) + (((uint64_t) (hi)) << 32))
+
+typedef int cryptonite_p256_err;
+typedef uint64_t cryptonite_p256_digit;
+typedef int64_t cryptonite_p256_sdigit;
+typedef __uint128_t cryptonite_p256_ddigit;
+typedef __int128_t cryptonite_p256_sddigit;
```
vincenthz, line 52: not a big problem, but I'ld imagine that could bite into portability

cbits/include64/p256/p256_gf.h
ocheron: Main goal here is to have a total of 257 bits just like the 32-bit version.
Some parts of the code depend on this, at least felem_is_zero_vartime.
Sticking to what already exists was best chance of completing something.

cbits/include64/p256/p256.h
ocheron: C wise it's certainly less portable, but already used in other parts of the project normally.

ocheron:
Thanks for giving a look.

PR #310, opened 2020-02-13
instance (MonadRandom m, MonadTrans t, Monad (t m)) => MonadRandom (t m)

### LINKED ISSUE #304: MonadRandom needs instances for transformers
Would you be OK with cryptonite incurring a dependency on transformers to provide instances for types in the transformers package (`instance MonadRandom {Except,Reader,Writer,...}T`)?

If not I might make a `cryptonite-orphans` package since that's handy and wouldn't step on any of your plans/toes.

frasertweedale:
It is common to use monad transformer stacks in applications, with
IO at the heart.  For applications using cryptonite, it is useful to
have MonadRandom instances for the common transformers.  Add the
instance

  instance (MonadRandom m, MonadTrans t, Monad (t m)) => MonadRandom (t m)

which achieves this with little code and one small additional
dependency, 'transformers', with no new transitive dependencies and
upon which many programs are likely to end up depending anyway.
Although new dependencies are discouraged, this change is motivated
by the usefulness of the instance and the desire to avoid orphan
instances in applications.

Also drive-by remove (Functor m) from MonadRandom class constraints,
because Functor is now a superclass of Monad and cryptonite no
longer supports pre-AMP GHC versions.

Fixes: https://github.com/haskell-crypto/cryptonite/issues/304

### DIFF
--- a/Crypto/Random/Types.hs
+++ b/Crypto/Random/Types.hs
@@ -5,6 +5,10 @@
 -- Stability   : experimental
 -- Portability : Good
 --
+
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+
 module Crypto.Random.Types
     (
       MonadRandom(..)
@@ -13,11 +17,12 @@ module Crypto.Random.Types
     , withDRG
     ) where
 
+import Control.Monad.Trans.Class
 import Crypto.Random.Entropy
 import Crypto.Internal.ByteArray
 
 -- | A monad constraint that allows to generate random bytes
-class (Functor m, Monad m) => MonadRandom m where
+class (Monad m) => MonadRandom m where
     getRandomBytes :: ByteArray byteArray => Int -> m byteArray
 
 -- | A Deterministic Random Generator (DRG) class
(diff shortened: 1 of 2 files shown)


ocheron:
What if I want:
```haskell
instance Monad m => MonadRandom (StateT ChaChaDRG m) where
    getRandomBytes n = StateT (return . randomBytesGenerate n)
```
Lifting is not the only valid implementation.
And an instance for `t m` would overlap with almost everything.

ocheron:
The goal of the typeclass is to abstract over different implementations.
We can't force an instance defined for all parameterized types.

frasertweedale:
@ocheron thanks for the consideration... and a justified rejection.

ocharles:
> What if I want:

> ```haskell
> instance Monad m => MonadRandom (StateT ChaChaDRG m) where
>    getRandomBytes n = StateT (return . randomBytesGenerate n)
> ```

AFAICT, this pull request does not stop you doing that. You simply need to mark your instance as overlapping. (Though typically you mark the general instance as overlappable, but it's the same idea).

ocheron:
Yes but it's not terribly good to rely on instance resolution when the impact of a change is so easily unnoticed.

Issue #212, opened 2018-01-12
Alternative ED25519 hash


rainbyte:
While learning about [RaiBlocks](https://github.com/clemahieu/raiblocks), I'm trying to replicate some of its functionality using Haskell.

As it is explained in their [design document](https://github.com/clemahieu/raiblocks/wiki/Design-features), ED25519 is used for key handling, but Blake2b is used instead of SHA-512 for hashing.

I have searched a way to select hashing type in Cryptonite, but it seems that SHA-512 is hardcoded in the C implementation (see [here](https://github.com/haskell-crypto/cryptonite/blob/master/cbits/ed25519/ed25519-hash.h)).

Maybe I'm mistaken, but if that is not the case, would it be possible to add a way to select Blake2b?

Thanks in advance!

P/D: I'm willing to contribute a patch if minimal instructions are given (it is my 1st time using Cryptonite)

ocheron:
With the Edwards25519 primitives it shouldn't be too difficult to reimplement Ed25519 in Haskell and have a parameter for the hash algorithm. And now even a constraint like `HashDigestSize hash ~ 64` is possible.

ocheron:
@rainbyte I recovered an implementation I did some time ago for testing:
https://gist.github.com/ocheron/892f4b50b7db5f4b3cebedf19d39835c

rainbyte:
Thanks for answering so fast! I will take a look at that code and try to adapt it

ocheron:
If something is required on the project I would go for something like this: in addition to modules Ed25519 and Ed448 we already have, a new module `Crypto.PubKey.EdDSA` with some features:
- selection of elliptic curve with a parameter (we have only Curve_Edwards25519 so far but Curve_Edwards448 may come later)
- control over hash algorithm to use internally
- support for context and prehash variants described in RFC 8032

Shou:
Glad I'm not alone here, I'm also trying to do something with Nano (Raiblocks) in Haskell. Is a patch still on the table? I'd also be willing to help out here. It'd probably also help to make the choice of hash function explicit in the Haddock documentation.

PR #312, opened 2020-02-24
Generic EdDSA implementation

ocheron:
My attempt at resolving #212 and replacing #298.

EdDSA is implemented in Haskell using Edwards25519 primitives, which allows to
select a different hash algorithm easily, and also add "context" and "prehash" variants.

Compared to C implementation in ed25519-donna there is a performance impact of up to 18%, visible when signing an empty message:

```
EdDSA/EdDSA-Ed25519/sign                 mean 21.09 μs  ( +- 97.52 ns  )
EdDSA/EdDSA-Ed25519/verify               mean 57.05 μs  ( +- 530.6 ns  )
EdDSA/Ed25519/sign                       mean 17.92 μs  ( +- 39.56 ns  )
EdDSA/Ed25519/verify                     mean 55.24 μs  ( +- 515.6 ns  )
```

The 3 μs cost is as follows:
- 1 μs because of ScrubbedBytes backend in scalar values. This could be decreased with a mutable scalar API, or no scrubbing at all.
- 1 μs because of hashing FFI calls, notably the hash "update" function which is not marked unsafe
- 1 μs as global overhead this approach has: additional Haskell allocations, indirections and FFI roundtrips


vincenthz:
this was some great work, and at least should be merged in.

Issue #314, opened 2020-03-03
blake2 hashes don't work on older CPUs if cryptonite was built with support_aesni


pvgoran:
I have a machine with an AMD Phenom II (`AMD Phenom(tm) II X3 720 Processor` as per `/proc/cpuinfo`).

When I build `cryptonite` on it without specifying any configuration options, it fails the Blake2 tests:

```
$ git clone https://github.com/haskell-crypto/cryptonite.git
$ cd cryptonite
$ cabal build -j
$ cabal test -j --test-options='-p Blake2'
...
Running 1 test suites...
Test suite test-cryptonite: RUNNING...
cryptonite
  hash
    KATs
      Blake2b-160
        1:         test-cryptonite: too many pending signals
Test suite test-cryptonite: FAIL
Test suite logged to:
/home/paul/build/cryptonite/dist-newstyle/build/x86_64-linux/ghc-8.6.5/cryptonite-0.26/t/test-cryptonite/test/cryptonite-0.26-test-cryptonite.log
0 of 1 test suites (0 of 1 test cases) passed.
cabal: Tests failed for test:test-cryptonite from cryptonite-0.26.
```

Also, when I build `git-annex` (https://git-annex.branchable.com/, uses cryptonite) and run its tests, they also fail, with a more specific message:

```
$ git clone git://git-annex.branchable.com/ git-annex
$ cd git-annex
$ cabal install -j --only-dependencies
$ cabal configure
$ cabal build -j
$ ./dist-newstyle/build/x86_64-linux/ghc-8.6.5/git-annex-8.20200226/build/git-annex/git-annex test -v -p Tests.QuickCheck.blake2s_160
Illegal instruction (core dumped)
```

(See https://git-annex.branchable.com/bugs/__34__Illegal_instruction__34___when_running_certain_tests_on_AMD_Phenom_II for where this came from.)

When I disable support_aesni by passing `--constraint="cryptonite -support_aesni"` to `cabal`, both of these tests pass.

This means that AES-NI is used in Blake2 implementation based on purely compile-time options. Because of this, any binary distribution of `cryptonite` (or a software which employs `cryptonite` for Blake2 hashes) will be crippled: if built with AES-NI support, it will fail hard on older CPUs; if built without AES-NI support, it will be slowed than it could be on newer CPUs.

Ideally, AES-NI should only be used after checking that it is actually available at runtime. Barring this, `support_aesni` should probably be disabled by default.

ocheron:
[...]

PR #316, opened 2020-03-15
AESNI and PCLMUL as per-function attributes

ocheron:
In #314 was confirmed that flag support_aesni causes runtime failures in code not related to AES-NI with i386/amd64 processor not supporting SSSE3. This comes from the use of option `-mssse3` at compile time, enabled for all C source files.

The PR adds an optional package flag to use instead function attributes, enabling the necessary target options where necessary, which means only in parts of the code protected with runtime CPU detection. The flag can be enabled in distributions when wanting broad CPU support and knowing the C compiler is compatible.

Mechanism known to be supported with:
- GCC >= 4.4 (released in 2009, see [changes](https://gcc.gnu.org/gcc-4.4/changes.html))
- clang >= 3.7 (released in 2015, see [docs](https://releases.llvm.org/3.7.0/tools/clang/docs/AttributeReference.html#target-gnu-target))

Related to:
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=869096
- https://github.com/NixOS/nixpkgs/issues/81915
- https://github.com/haskell/cabal/issues/4294

### DIFF
--- a/cryptonite.cabal
+++ b/cryptonite.cabal
@@ -105,7 +105,7 @@ Flag check_alignment
 
 Flag use_target_attributes
   Description:       use GCC / clang function attributes instead of global target options.
-  Default:           False
+  Default:           True
   Manual:            True
 
 Library
--- a/cryptonite.cabal
+++ b/cryptonite.cabal
@@ -103,6 +103,11 @@ Flag check_alignment
   Default:           False
   Manual:            True
 
+Flag use_target_attributes
+  Description:       use GCC / clang function attributes instead of global target options.
+  Default:           False
+  Manual:            True
+
 Library
   Exposed-modules:   Crypto.Cipher.AES
                      Crypto.Cipher.AESGCMSIV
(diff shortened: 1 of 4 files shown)


cryptonite.cabal
```diff
@@ -103,6 +103,11 @@ Flag check_alignment
   Default:           False
   Manual:            True
 
+Flag use_target_attributes
+  Description:       use GCC / clang function attributes instead of global target options.
+  Default:           False
```
vincenthz: I'm tempted to have this default to `true` here and let people with old compiler have the pain & flag to fallback to the old behavior ?

ocheron:
Flag is enabled by default.

Version requirement with GCC is not high.

For clang the requirement is similar to LLVM version for GHC itself. According to this table most impacted users would be OS X 10.10 Yosemite with Xcode 6, but Xcode 7 is also available: https://trac.macports.org/wiki/XcodeVersionInfo

cryptonite.cabal
ocheron: Thanks for the push, I was tempted too.

Issue #302, opened 2019-12-02
Serialization in compressed format


gromakovsky:
In my application I need to encode and decode `secp256k1` public keys in the same format as Bitcoin uses: https://www.oreilly.com/library/view/programming-bitcoin/9781492031482/ch04.html

According to that page it is called "Compressed SEC Format". I know that there is another library specialized to `secp256k1`, but I am already using `cryptonite` and don't want to add more dependencies. I am using `Crypto.PubKey.ECC.ECDSA.PublicKey` with hardcoded `Curve`, so I only need to encode and decode `Crypto.PubKey.ECC.Types.Point`. It is quite easy to implement encoding function (here I am using `ByteString` for simplicity, it can be generalized to `ByteArray`):
```
publicKeyToBytes :: HasCallStack => ECDSA.PublicKey -> ByteString
publicKeyToBytes (ECDSA.PublicKey curve publicPoint) =
  case publicPoint of
    -- We are using `i2ospOf_` because `curveSizeBits` ensures that
    -- the number won't have more than that many bytes.
    Point x y -> prefix y <> i2ospOf_ (curveSizeBits curve `div` 8) x
    PointO -> error "PublicKey somehow contains infinity point"
  where
    prefix :: Integer -> ByteString
    prefix y
      | odd y = "\003"
      | otherwise = "\002"
```

Decoding function is a bit trickier, I haven't tried implementing it yet, but the algorithm is described in the linked page, so I think I can do it as well.

I know that there is `EllipticCurve` class with `encodePoint` and `decodePoint` functions, but I couldn't find any implementation that uses compressed format.

So I have a couple of questions:
1. Is such functionality already present anywhere in the library (maybe I just missed that)? Maybe not encoding/decoding, but at least a function which takes `x` coordinate and `Curve` and returns `y` coordinate or `Point`? It would be very helpful for decoding.
2. If there is no such functionality, does it make sense to add it? If yes, where should I put it? I can do it in `Crypto.PubKey.ECC.ECDSA` or I can add a newtype wrapper which customizes `EllipticCurve` instance so that `encodePoint` and `decodePoint` use compressed format.

P. S. I am a crypto newbie, so please don't hate me if I wrote nonsense ^_^

ocheron:
We don't have this currently. Compressed format saves some space but the deserialization is very slow compared to just validation of (x,y).

```
EC/SEC_p256k1/uncompressed               mean 650.1 ns  ( +- 18.28 ns  )
[...]

PR #322, opened 2020-06-02
implement square roots in f2m

incertia:
this should address the final checkbox in #302. quadratic equation doesn't seem to fit the theme of the module so it is left as an exercise to the reader. i think the haddock is good enough and i cleaned up some error strings in the f2m module as well.

we perform the tests in GF(2^16) because it is an exercise to find an irreducible polynomial of given degree and a random number doesn't cut it.

### DIFF
--- a/Crypto/Number/F2m.hs
+++ b/Crypto/Number/F2m.hs
@@ -17,7 +17,6 @@ module Crypto.Number.F2m
     , squareF2m'
     , squareF2m
     , powF2m
-    , powF2m'
     , modF2m
     , sqrtF2m
     , invF2m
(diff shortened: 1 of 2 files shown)


ocheron:
The module doesn't need to expose both `powF2m` and `powF2m'`.
Only one function is enough.

EDIT: I see the goal with `powF2m` was to eliminate Maybe

Also `sqrtF2m` don't need to go through the full `powF2m` with a large `2 ^ (log2 fx - 1)` argument. It simplifies to repeated call to `squareF2m` with a (small) loop counter.

incertia:
updated `sqrtF2m`

ocheron:
Thank you for the contribution.
I removed `powF2m'` as I don't know if it's so useful. With our modular exponentation functions, not everything supports negative exponents.
And fixed `powF2m` to use field multiplication.

Issue #15, opened 2015-06-15
`Crypto.Random` and forking the `DRG`


ibotty:
Am I right, that forking the DRG is just not possible with every `DRG` in a pure way (i.e. without mixing entropy or using new entropy)?

If so, can you comment on how you envision it to be done? Maybe provide the building blocks to "purely" (but not referential transparent) mix entropy to get two generators or (better) provide a function to do that.

I might not have thought this out though.

vincenthz:
yes, it's not currently possible for the DRG class itself to fork/split.

You could very easily have a main "seed" generator that would seed temporary generators. i.e. something like this:

``` haskell
let (newDRG, seedGenerator') = withDRG seedGenerator drgNew
```

If we had a generic 'drgNew' for the DRG class, then you'ld be able to do that for every DRG.
Would that be enough to cover your cases ?

ibotty:
Yes, a `drgNew` (or `newDRG`) would be enough.

That said, after porting the code that triggered that issue to use the 
`MonadRandom` class, I don't seem to need it anymore.

PR #323, opened 2020-06-08
DRG interface improvements

infinity0:
This PR adds a `PRG` typeclass that supports initialising from a seed (which the system DRG cannot implement), and exposes raw functionality in ChaCha so other people can implement their own PRGs.

Motivation: I'm writing a network simulator that can run simulations deterministically with arbitrary suspend-resume capabilities. To do this, I need to be able to suspend the state of the PRG and write it to disk, then read it later. I appreciate this might be a controversial feature so didn't include it in this PR, but have merely exposed the functionality needed to easily write my own PRG.

If you like, I can also add a commit that includes a `ChaChaPureInsecure` instance of `PRG` that uses `ByteString` and derives `(Show, Read, Generic)`, etc for serialisation purposes. (Using `ScrubbedBytes` is also possible for more security, but is more work as you have to write some of the instances by hand, I didn't get around to it yet.)


ocheron:
For now I looked only at the ChaCha changes.
You generalize `StateSimple` to arbitrary `ByteArray` instances.
This has 2 consequences:
(1) It can't ensure that the input to `generateSimple` has expected length, so the C code may read past the buffer end
(2) It exposes state internals which are architecture dependent

While for (1) the class in `Data.ByteArray.Sized` could help, for (2) maybe explicit functions to serialize/deserialize `StateSimple` in a format with defined and stable endianness would be preferable.

infinity0:
@ocheron Thanks for the comment, I overlooked the arch-dependent-ness of it. How about 75d78763 ?

infinity0:
@ocheron Hi, mind taking another look?

Crypto/Cipher/ChaCha.hs
```diff
@@ -33,6 +37,19 @@ newtype State = State ScrubbedBytes
...
+-- Note: anyone that can read this string, can break the security of the RNG.
+-- Treat it with as much importance as you would a password.
+toPortable :: StateSimple -> BS.ByteString
+toPortable (StateSimple st) = unsafeMapWords toLE32 st
+
+-- | Convert a 'StateSimple' from a little-endian 'BS.ByteString' into its
+-- internal architecture-dependent representation.
+fromPortable :: BS.ByteString -> StateSimple
+fromPortable bs = StateSimple $ unsafeMapWords fromLE32 bs
+
```
ocheron, line 56: fromPortable should verify input length and return a CryptoFailable StateSimple (for example with a new error CryptoError_StateSizeInvalid). This is how we do it elsewhere, for example Curve25519.publicKey.

Crypto/Internal/ByteArray.hs
```diff
@@ -37,3 +40,17 @@ constAllZero b = unsafeDoIO $ withByteArray b $ \p -> loop p 0 0
...
+                . (ByteArrayAccess ba1, ByteArray ba2, Storable a)
+               => (a -> a) -> ba1 -> ba2
+unsafeMapWords f prevSt = unsafeDoIO $ do
+    newSt  <- copy prevSt (\_ -> return ())
+    withByteArray newSt $ \stPtr -> do
+        let sz = Data.ByteArray.length prevSt `div` sizeOf (undefined :: a)
+        for_ [0 .. pred sz] $ \i -> do
+            w <- peekElemOff stPtr i
+            pokeElemOff stPtr i (f w)
+    return newSt
```
ocheron: A single call to copyAndFreeze can replace the 3 calls to unsafeDoIO, copy, withByteArray.

Crypto/Random/Types.hs
```diff
@@ -5,16 +5,21 @@
...
 --
+{-# LANGUAGE ScopedTypeVariables #-}
 module Crypto.Random.Types
     (
       MonadRandom(..)
     , MonadPseudoRandom
     , DRG(..)
+    , PRG(..)
+    , prgNewEntropy
+    , prgNew
```
ocheron, line 15: I would prefer not to expose prgNewEntropy unless absolutely necessary. This function relies on its argument always returning requested length.

Crypto/Random/Types.hs
```diff
@@ -25,6 +30,26 @@ class DRG gen where
...
     randomBytesGenerate :: ByteArray byteArray => Int -> gen -> (byteArray, gen)
 
+-- | A Psuedo Random Generator (PRG) class
+--
+-- Like 'DRG' but also supports initialisation from some fixed seed.
+class DRG gen => PRG gen where
+    -- | Initialize the DRG from some fixed seed.
+    prgNewSeed :: ByteArrayAccess seed => seed -> gen
+    -- | Length of seed in bytes
+    prgSeedLength :: Proxy gen -> Int
```
ocheron: prgSeedLength could use a type variable `proxy` instead of concrete `Proxy`, this is more general, and ensures even better that the argument is never pattern-matched.

ocheron (commented):
I'm not super familiar with the random part but think this is ok to add, with a few review comments.

Crypto/Random/Types.hs
infinity0: Done in 1686aee

Crypto/Cipher/ChaCha.hs
infinity0: Done in 1686aee

Crypto/Internal/ByteArray.hs
infinity0: Done in 1686aee

Crypto/Random/Types.hs
infinity0: I can remove it, but it's really just a convenience wrapper around `prgNewSeed`, a typeclass method of the new `PRG` class I'm adding in this PR. Anyone can recreate `prgNewEntropy` themselves, so I'm not sure hiding it achieves anything in terms of avoiding misuse.

For `prgNewSeed`, I basically copied the type signature of the existing `ChaCha.initializeSimple`. Its implementation checks the length of the seed, but the type signature does not make this obvious. So I could modify both of these to return a `CryptoFailable StateSimple/gen` instead - and then this type would propagate through into `prgNewEntropy`. How does that sound? That should make it much more obvious that anyone implementing `prgNewSeed` should explicitly check the length.

infinity0: For example, in 992207539907d343b2d6435da1407e96b6353fda

ocheron: I still think the real convenience wrapper is `prgNew`, without a `CryptoFailable` result.
`MonadRandom` is supposed to always return the requested length.
So if the `PRG` and `MonadRandom` instances are implemented correctly, a result _will_ be generated.

Misuse comes only with `prgNewEntropy`.
People not wanting `getRandomBytes`, or wanting length larger than minimum, can call the `PRG` methods directly.
They will know best if the length requirement is always met or not.

infinity0: In the interests of getting this PR merged, I don't want to continue arguing over a minor point, and so I've un-exposed `prgNewEntropy` as you wanted, and made `prgNew` auto-unwrap the `CryptoFailable`.

However, I still believe the old version is OK and more helpful for users. In Haskell, typeclass instances are chosen by the caller, so I don't see a fundamental usage/abuse/security difference between:

1. letting the user implicitly select the instance of `MonadRandom` class in `prgNew`, vs
2. letting the user explicitly select the `getEntropy` function in `prgNewEntropy`

Either of these could return the wrong length, but (a) the programmer can easily check the function they're using behaves correctly and (b) since 9922075 we mandate a length check in `prgNewSeed`.

Arguably, the explicit version (2) is easier to check - with `MonadRandom` (1) you have to look up the instance definition first, which could be a redirection to another function (e.g. `MonadRandom IO`), so it takes two steps. This is why in my own code, I prefer to use `getEntropy` explicitly and avoid the implicit `MonadRandom IO` instance.

infinity0:
ping @ocheron

Crypto/Random/Types.hs
ocheron: I don't want to merge `initializeSimpleErr` like this.
Bytes after the 40th are ignored and don't contribute to the DRG.

infinity0: @ocheron It is the same as the current behaviour of `initializeSimple` - as you can see in the diff I did not change the behaviour. How do you want to fix it, and shouldn't it be done in a separate PR?

infinity0:
ping @ocheron I am still interested in getting this PR merged, can you elaborate on the next steps?

ocheron:
Sorry I don't do Haskell anymore and have left the project.

infinity0:
@ocheron OK, thanks for the response and best of luck with whatever you're doing now!

@vincenthz can you take over and have a look please?

infinity0:
Ping @vincenthz again - I've also added a generic `prgFork` implementation that addresses #15.

vincenthz:
archiving repository

PR #331, opened 2020-08-07
Hash data in 4GB chunks to avoid uint32_t overflow.

### LINKED ISSUE #330: Incorrect hashes for all hash algorithms beyond 4 GiB of input
I found a problem together with @chpatrick:

<details>

<summary>Click to show imports</summary>

```
:set -XTypeApplications
import qualified Crypto.Hash as Hash
import qualified Data.ByteString as BS
import qualified Data.ByteString.Char8 as BS8
```

</details>

```
> let hash = show . Hash.hash @BS.ByteString @Hash.SHA1
> let a = BS8.replicate 5000000000 'a' <> BS8.pack "hello"
> let b = BS8.replicate 5000000000 'a' <> BS8.pack "world"
> a == b
False
> hash a == hash b
True
```

**Two different ByteStrings hash to the same SHA1.**

Trying SHA256, same issue:

```
> let hash = show . Hash.hash @BS.ByteString @Hash.SHA256
> hash a == hash b
True
```

# Conclusion

* `fromIntegral` is bad.
  Until we deprecate `fromIntegral`, **Haskell code will always be subtly wrong and never be secure.**
  I [proposed](https://mail.haskell.org/pipermail/libraries/2017-September/028216.html) this 3 years ago (easier-to-read threaded view [here](http://haskell.1045720.n5.nabble.com/Deprecating-fromIntegral-tt5864578.html)).
  Concretely:
    * If we don't fix this, people will shy away from using Haskell for serious work. I found this bug in production, and you can imagine how happy I am about that. Rust and C both do this better.
    * If the authors of key crypto libraries fall for these traps (no blame on them), who can get it right? We should remove the traps.

# Details

## Demonstrating that the the overflow occurs at the `uint32` boundary:

Good:

```
> show $ Hash.hash @BS.ByteString @Hash.SHA1 (BS8.replicate (2^32 - 2) 'a' <> BS8.pack "b")
"78c1411ef0ae4e0328660b7dddc08c06000eb9f0"

> show $ Hash.hash @BS.ByteString @Hash.SHA1 (BS8.replicate (2^32 - 2) 'a' <> BS8.pack "c")
"21f556aba357dd19d7bf01a2a7f05482162ed3fc"
```

Bad:

```
> show $ Hash.hash @BS.ByteString @Hash.SHA1 (BS8.replicate (2^32 - 1) 'a' <> BS8.pack "b")
"da39a3ee5e6b4b0d3255bfef95601890afd80709"

> show $ Hash.hash @BS.ByteString @Hash.SHA1 (BS8.replicate (2^32 - 1) 'a' <> BS8.pack "c")
"da39a3ee5e6b4b0d3255bfef95601890afd80709"
```

## Location of the problem:

For SHA1 (note `uint32_t len`):

https://github.com/haskell-crypto/cryptonite/blob/10dc63c51f89e747a25e636de34d8ccc8fefbdbf/cbits/cryptonite_sha1.c#L159

For all hash algorithms (note `Word32` for the length parameter):

https://github.com/haskell-crypto/cryptonite/blob/10dc63c51f89e747a25e636de34d8ccc8fefbdbf/Crypto/Hash/Types.hs#L59-L60

This is where the **silent overflow occurs due to `fromIntegral`**:

https://github.com/haskell-crypto/cryptonite/blob/10dc63c51f89e747a25e636de34d8ccc8fefbdbf/Crypto/Hash.hs#L91

chpatrick:
This fixes #330.

I think `hashFinalizePrefix` could have the same issue if you give it a huge buffer though, and this doesn't fix it.

### DIFF
--- a/Crypto/Hash.hs
+++ b/Crypto/Hash.hs
@@ -50,11 +50,11 @@ import           Basement.Block.Mutable (copyFromPtr, new)
 import           Crypto.Internal.Compat (unsafeDoIO)
 import           Crypto.Hash.Types
 import           Crypto.Hash.Algorithms
-import           Foreign.Ptr (Ptr)
+import           Foreign.Ptr (Ptr, plusPtr)
 import           Crypto.Internal.ByteArray (ByteArrayAccess)
 import qualified Crypto.Internal.ByteArray as B
 import qualified Data.ByteString.Lazy as L
-import           Data.Word (Word8)
+import           Data.Word (Word8, Word32)
 
 -- | Hash a strict bytestring into a digest.
 hash :: (ByteArrayAccess ba, HashAlgorithm a) => ba -> Digest a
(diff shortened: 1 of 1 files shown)


vincenthz:
Thanks, that should do the job, but still cannot recommend of using the >4GB strict bytestring.

NorfairKing:
This is being fixed without a regression test. What prevents this from being reintroduced?

nh2:
> What prevents this from being reintroduced?

Nothing; we also don't know whether this completely fixes it everywhere (see PR description). We just upstreamed the fix that makes those functions that we use work.

Issue #255, opened 2018-10-24
Add an entropy source for Linux `getrandom`


taktoa:
On Linux, `/dev/urandom` is unsafe in early boot as it will not block even if an insufficient amount of entropy has been gathered. On the other hand, `/dev/random` is complete overkill, as it will block even if a sizeable entropy pool has been gathered -- we should usually trust CSPRNGs as long as they have a decent amount of fresh entropy to work with. On Linux, our best bet to avoid this issue is the [`getrandom`](http://man7.org/linux/man-pages/man2/getrandom.2.html) syscall, which has exactly the right semantics: it blocks if `/dev/urandom` hasn't been initialized yet, and then provides randomness the same way `/dev/urandom` would normally.

PR #339, opened 2021-02-18
Add Entropy source using getrandom()

eamsden:
This PR addresses #255 by implementing getrandom and, depending on a flag, replacing `/dev/random` and `/dev/urandom` as entropy sources with the `getrandom()` syscall.

`getrandom()` is strictly better than /dev/random or /dev/urandom. It will block on systems with uninitialized entropy pools, but will not block thereafter, and pulls from the same entropy pool as `/dev/urandom`. Thus, it does not block after system entropy initialization, like `/dev/random` does, and it does not provide "bad" entropy prior to system entropy initialization, like `/dev/urandom` does.

It is also not vulnerable to file descriptor exhaustion, as it does not use filehandles to read from the stream devices, but instead generates bytes directly from the kernel entropy pool and copies them into the provided buffer.

### DIFF
--- a/Crypto/Random/Entropy/Backend.hs
+++ b/Crypto/Random/Entropy/Backend.hs
@@ -22,6 +22,8 @@ import Crypto.Random.Entropy.RDRand
 #endif
 #ifdef WINDOWS
 import Crypto.Random.Entropy.Windows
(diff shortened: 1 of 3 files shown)


vincenthz:
archiving repository

PR #345, opened 2021-04-14
Fix for 32 bit platforms

hamishmack:
The use of `(fromIntegral (maxBound :: Word32))` causes problems.  It is used to make an `Int` and 32 bit systems it winds up being -1.


vincenthz:
Thanks, that makes sense for the 32 bits platform. Seems a bit excessive to pull Int32 just to get it's max bound, maybe more direct to just choose a chunk size directly like 1gb or 2gb and put the value in hexadecimal ?

PR #349, opened 2021-09-13
Linker complaints

iquerejeta:
Not all blake functions have the prefix, which results in linker errors. Just putting a `_cryptonite_` prefix to avoid such cases. Reason of choosing `_cryptonite_` over `cryptonite_` is because the exposed blake functions already have the `cryptonite_` prefix, which would be a bit of a mess (to have different functions with the same name).

### DIFF
--- a/cbits/blake2/ref/blake2.h
+++ b/cbits/blake2/ref/blake2.h
@@ -142,51 +142,51 @@ extern "C" {
   };
 
   /* Streaming API */
-  int blake2s_init( blake2s_state *S, size_t outlen );
-  int blake2s_init_key( blake2s_state *S, size_t outlen, const void *key, size_t keylen );
-  int blake2s_init_param( blake2s_state *S, const blake2s_param *P );
-  int blake2s_update( blake2s_state *S, const void *in, size_t inlen );
-  int blake2s_final( blake2s_state *S, void *out, size_t outlen );
+  int _cryptonite_blake2s_init( blake2s_state *S, size_t outlen );
+  int _cryptonite_blake2s_init_key( blake2s_state *S, size_t outlen, const void *key, size_t keylen );
+  int _cryptonite_blake2s_init_param( blake2s_state *S, const blake2s_param *P );
+  int _cryptonite_blake2s_update( blake2s_state *S, const void *in, size_t inlen );
+  int _cryptonite_blake2s_final( blake2s_state *S, void *out, size_t outlen );
 
-  int blake2b_init( blake2b_state *S, size_t outlen );
-  int blake2b_init_key( blake2b_state *S, size_t outlen, const void *key, size_t keylen );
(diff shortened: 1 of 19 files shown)


cbits/blake2/ref/blake2.h
```diff
@@ -142,51 +142,51 @@ extern "C" {
...
-  int blake2bp_init_key( blake2bp_state *S, size_t outlen, const void *key, size_t keylen );
-  int blake2bp_update( blake2bp_state *S, const void *in, size_t inlen );
-  int blake2bp_final( blake2bp_state *S, void *out, size_t outlen );
+  int _cryptonite_blake2s_init( blake2s_state *S, size_t outlen );
+  int _cryptonite_blake2s_init_key( blake2s_state *S, size_t outlen, const void *key, size_t keylen );
+  int _cryptonite_blake2s_init_param( blake2s_state *S, const blake2s_param *P );
+  int _cryptonite_blake2s_update( blake2s_state *S, const void *in, size_t inlen );
+  int _cryptonite_blake2s_final( blake2s_state *S, void *out, size_t outlen );
+
+  int _cryptonite__cryptonite_blake2b_init( blake2b_state *S, size_t outlen );
```
vincenthz: this one suffers from double substitution

vincenthz (changes requested):
LGTM, just one symbol with 2 subtitutions

cbits/blake2/ref/blake2.h
iquerejeta: changed that, and remove the prefix to some static functions. Couldn't test the correct compilation of ref though. Any suggestions how to test that? Thanks

cbits/blake2/ref/blake2b-ref.c
```diff
@@ -95,11 +95,11 @@ int blake2b_init_param( blake2b_state *S, const blake2b_param *P )
 
 
 
-int blake2b_init( blake2b_state *S, size_t outlen )
+int _cryptonite_blake2b_init( blake2b_state *S, size_t outlen )
 {
   blake2b_param P[1];
 
-  if ( ( !outlen ) || ( outlen > BLAKE2B_OUTBYTES ) ) return -1;
+  if ( ( !outlen ) || ( outlen > blake2B_OUTBYTES ) ) return -1;
```
vincenthz: decapitalized by mistake ?

vincenthz (approved):
LGTM, just need fixing the constant value and we can merge

cbits/blake2/ref/blake2.h
vincenthz: only way is to do it manually in the .cabal file:

```diff
-  if arch(x86_64) || flag(support_sse)
+  if false -- arch(x86_64) || flag(support_sse)
```

vincenthz:
thanks, merged after fixing the 2 last renaming issues in the the bp and sp versions

PR #351, opened 2021-09-30
Add RandomGen instance for ChaChaDRG

kephas:


vincenthz:
As per the contribution guidelines:
https://github.com/haskell-crypto/cryptonite/blob/master/CONTRIBUTING.md

I don't see a need to add `random`, and it could be harmful to have this instance also (confusion related to the cryptographic "quality" of the random generator). I think the best option if you want to do this, is providing your own newtype instance, which should be very similar to this patch

PR #353, opened 2021-10-07
Add Show and Data instances to HMAC

eyeinsky:
(Derived via `Digest a` which the HMAC newtype wraps and which already has these instances.)

### DIFF
--- a/Crypto/MAC/HMAC.hs
+++ b/Crypto/MAC/HMAC.hs
@@ -10,6 +10,7 @@
 --
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE DeriveDataTypeable #-}
 module Crypto.MAC.HMAC
     ( hmac
     , hmacLazy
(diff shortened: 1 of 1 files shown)


Crypto/MAC/HMAC.hs
```diff
@@ -30,13 +31,14 @@ import qualified Crypto.Internal.ByteArray as B
...
 import qualified Data.ByteString.Lazy as L
+import           Data.Data (Data)
 
 -- | Represent an HMAC that is a phantom type with the hash used to produce the mac.
 --
 -- The Eq instance is constant time.  No Show instance is provided, to avoid
 -- printing by mistake.
 newtype HMAC a = HMAC { hmacGetDigest :: Digest a }
-    deriving (ByteArrayAccess)
+    deriving (ByteArrayAccess, Show, Data)
```
vincenthz, line 41: the comment above :
```
No Show instance is provided, to avoid printing by mistake.
```

I'm ok with `Data`, but `Show` still doesn't seems like a good idea

eyeinsky: Right, missed the comment part. But two questions then: what is the purpose of Show (why is it bad to print it)? and also, the upstream Digest has Show, why is it ok there?

My purpose for Show is pretty much only debugging, i.e not serialization to later read it back in later.

vincenthz: HMAC is usually intended as a secret value, which is why the `==` is constant time .. so printing by mistake could have consequence, for example writing an HMAC in the log. Which is why, it's somewhat safer to have an explicit step to get the digest "inside" which you can print (`hmacGetDigest`).

eyeinsky: Hmm, I'm probably missing something here, but isn't the HMAC itself public? You have a secret and a message to create the HMAC, which you can pass to somewhere untrusted, and then get it back and verify that the message is still authentic.

Or is the HMAC lazy, printing it evaluates it and then the compare speed (using the constant time Eq) would change?

vincenthz:
archiving repository

PR #362, opened 2022-03-22
Add optional MTL support

matobet:
Ease usage of `MonadRandom` in monad transformer stacks.

Projects using MTL will benefit from being able to automatically derive `MonadRandom` instance over their custom monad transformers. 

Added new cabal flag `support_mtl` (default `false`) so users that are not interested in MTL wouldn't have to suffer from extra dependency.


matobet:
As an additional motivation, we already see people hand-rolling their own orphans like this in the ecosystem https://github.com/frasertweedale/hs-jose/blob/master/src/Crypto/JOSE/Error.hs#L121 
Would be better to have it as part of the main package.

vincenthz:
archiving repository

PR #363, opened 2022-03-31
add rawSHAKE

cielavenir:
rawSHAKE is mentioned in https://github.com/maandree/libkeccak/blob/master/libkeccak.h#L41 and https://github.com/maandree/sha3sum


vincenthz:
archiving repository

PR #365, opened 2022-04-26
Add support for XChaCha and XChaChaPoly1305

MatthiasValvekens:
This PR adds support for the extended nonce variants of ChaCha and ChaChaPoly1305, as implemented in libsodium (among others), and specified in this RFC draft: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-xchacha-03. Like #364, I'm submitting this PR because I'd like to have this functionality in another project of mine that depends on `cryptonite`.

Since the only difference with the "base" cipher lies in the way the initial state is generated, the only change(s) in the Haskell API are a couple new initialisation functions & associated types. I considered overloading the existing `initialize` functions and branching on the length of the nonce, but I figured keeping them separate would be cleaner.

This addition did require a number of additions at the C level, though (mainly to implement the extra HChaCha primitive). I haven't written any "serious" C in a while, so some extra attention would be appreciated there. Obviously, comments on other aspects of the implementation are also more than welcome.

Thanks a lot!


vincenthz:
thanks for the contribution, that looks useful and looks good, but you need to fix the indentation of the C files are mixed, make sure your editor default to tab / noexpand

MatthiasValvekens:
Oh, good catch. I didn't notice, and my editor is indeed set to expand all tabs. My bad, I'll fix that tonight. Thanks!

MatthiasValvekens:
Hi @vincenthz, I converted the indentation in the C files to use tabs instead of spaces. I preserved spaces where used for alignment rather than indentation, though (in this case: function signatures spanning multiple lines).

If there's anything else I need to change, please let me know. Thanks!


PS: I checked my other PR #364 as well, just to make sure. That one seems to be OK as far as indentation is concerned.

vincenthz:
archiving repository

PR #369, opened 2022-08-01
Update DECAF to fix AVX2 builds

mklca:
* based on v1.0.0 at https://sourceforge.net/projects/ed448goldilocks/
* add x86_64 architecture
* use vector alignment macro on struct rather than constant (16, needed
  to be 32 on AVX2 or else we break alignment assumptions).


mklca:
Incidentally, I put the nix infrastructure that builds this package with a fully Skylake architecture-enabled toolchain on [another branch](https://github.com/VitalBio/cryptonite/tree/avx2-fixes). (Edit: Fixed link)

cbits/decaf/p448/f_field.h
```diff
@@ -27,7 +27,7 @@
 #define SER_BYTES 56
 typedef struct cryptonite_gf_448_s {
     word_t limb[NLIMBS];
-} __attribute__((aligned(16))) cryptonite_gf_448_s, cryptonite_gf_448_t[1];
+} VECTOR_ALIGNED cryptonite_gf_448_s, cryptonite_gf_448_t[1];
```
mklca, line 30: Observed a general protection fault at `vmovdqa (%r15),%ymm2` when `r15` held the value `0x7fffffff53d0`. 16 byte alignment is not sufficient for AVX2 (needs to be 32).

cryptonite.cabal
```diff
@@ -298,28 +299,40 @@ Library
   else
     include-dirs:      cbits/include32
 
-  if arch(x86_64) || arch(aarch64)
-    C-sources:         cbits/decaf/p448/arch_ref64/f_impl.c
+  if arch(x86_64)
```
mklca, line 302: This nested `if` is required since the Cabal version floor for this package is currently 1.8.

mklca:
Have you fine folks (haskell-crypto team members @vincenthz @tekul @centromere) had a chance to look over this yet?

vincenthz:
archiving repository

